Issue #11, opened 2023-03-02
Add `vitest` setup for testing


BracketJohn:
## Goal

Add `vitest` to this code-base to implement tests

## Tasks

- [ ] add vitest
- [ ] set a minimum testing coverage 
- [ ] write the first dummy-test

PR #13, opened 2023-03-02
Feat: Add Vitest

KirillDogadin-std:
closes https://github.com/makerdao-ses/switchboard-boilerplate/issues/11


README.md
```diff
@@ -20,7 +20,11 @@ npm run dev
 # Typechecking (via TypeScript / tsc)
 npm run typecheck
 
-# Linting (via eslint)
```
BracketJohn, line 23: what happened to eslint doc? 😢

vitest.config.ts
```diff
@@ -0,0 +1,14 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    coverage: {
+      provider: "istanbul",
+      reporter: ["text", "json", "html"],
```
BracketJohn, line 7: what does this mean? If its the kind of file formatas that are supported: why do we not need to specify js / ts?

BracketJohn (changes requested):
Please also:
- [ ] add a ci step that runs tests
- [ ] enforce coverage of 90 percent (if possible at the moment)

README.md
KirillDogadin-std: i thought the command was removed but someone forgot to remove the comment as well

vitest.config.ts
KirillDogadin-std:
```
npx vitest run --coverage --reporter=json
```

to see the json formatted output. but seems to work without this.

README.md
```diff
@@ -21,6 +21,13 @@ npm run dev
 npm run typecheck
 
 # Linting (via eslint)
+npm run lint
+
+# Testing via vitest
+npm run test
+
+# Coverage via vitest, istambul coverage provider
+npm run test:cov
```
BracketJohn, line 30: let's turn on coverage by default!

BracketJohn (commented):
you ignored my general review comment from last time

README.md
BracketJohn: is coverage enforced by default now? I thought to rememvber that one has to turn in on

BracketJohn:
test cov is not working: Right now it shows 0 LoC, but this is not true -> it should find the src/* code and detect that 0 percent of it are covered.

BracketJohn (approved):

Issue #2, opened 2023-02-15
Research existing solutions


valiafetisov:
### Goal

Overview on the existing tools that we can use

### Context

During the initial planning, we raised the questions and mentioned few existing solutions which might cover some of the requirements. Those include [sanity](https://www.sanity.io/graphql), [strapi](https://strapi.io/), [slicknode](https://github.com/slicknode/slicknode) for headless CMSs and [postgraphile](https://github.com/graphile/postgraphile) and [graphile](https://www.graphile.org/) for automatically generated graphql endpoints.

### Tasks

- [x] Evaluate existing tools, initially proposed and newly discovered
  - [x] Check their applicability to the raised requirements
  - [x] Check their licences
- [ ] Conclude the very initial technical stack

valiafetisov:
### Headless CMSs

- [**Sanity**](https://www.sanity.io/graphql) – "GraphQL Headless CMS"
    - [MIT License](https://github.com/sanity-io/sanity/blob/next/LICENSE)
    - Cons
        - Only the [frontend/app](https://github.com/sanity-io/sanity) is open-source, but backend is not
        - They doesn't even advertise what database they use, seems to be graph-based
- [**Strapi**](https://github.com/strapi/strapi) – "open-source headless CMS"
    - [MIT Expat License](https://github.com/strapi/strapi/blob/main/LICENSE)
    - Available as completely self-hosted "community edition"
        - [Getting-started page](https://docs.strapi.io/developer-docs/latest/getting-started/quick-start.html)
    - Pros
        - ✔️ Supports `sqlite` or `postgres` or `mysql` as the data storage
        - ✔️ Have REST as well as graphql API endpoints out of the box
        - ✔️ Have API token generation with granular roles available
        - `+/-` Have user roles out of the box
            - But they are limited to 3 default one (that's their main selling feature for the paid version) – I think workaround/extensions are possible, since the source code is available
        - ✔️ UI for creating, editing, deleting of records (+ optional draft/publication of records)
        - Other
            - (optional) UI for creating, editing and deleting tables (available only to superadmins), although the change is stored as a code change, not in the database! So git-based review process is required to modify database structure and the models
                - it is possible to extend models with custom logic
                - [Migrations are possible](https://docs.strapi.io/developer-docs/latest/developer-resources/database-migrations.html) but are using `knex` and written in js
            - Webhooks on content edit event
            - File upload (to AWS S3 / locally)
            - Custom plugin support
[...]

PR #16, opened 2023-03-10
Add user + auth

### LINKED ISSUE #10: Implement the Auth.js database model and its module
## Goal

Implement the Auth.js database model and its module

## Context

Depends on #9. We want to first implement the auth.js module and its related logic to get a feeling for it.

## Tasks

- [ ] add the User db schema from https://github.com/makerdao-ses/ecosystem-api/tree/main/src/modules/Auth to the prisma schema
- [ ] add the module logic from https://github.com/makerdao-ses/ecosystem-api/tree/main/src/modules/Auth module to this project
- [ ] open a PR to discuss what it looks like

KirillDogadin-std:
close #10 

0. Run the service
    - `cp developer.env .env`
    - `npx prisma db push`
    - `npm run dev`
1. Go to `localhost:3000`
2. Use the following mutation to create user
```gql
mutation {
  signUp(user: { username: "asdf", password: "asdf" }) {
    user {
      username
    }
    token
  }
}
```
3. Use the following mutation to login
```gql
mutation {
  signIn(user: { username: "asdf", password: "asdf" }) {
    token
    user {
      username
    }
  }
}
```
4. see the user at the prisma playground.
    - run `npx prisma studio`
    - browser pops up
    - go to user table via ui and find your user there.
5. now you can query the `me` endpoint and make sure that being logged in allows you to access it:

add the following to the "Headers" menu (at the bottom left side of the tool):
```
{
  "Authorization": "Bearer <your token here you got from signIn endpoint>"
}
```

run this

```gql
query {
  me {
    username 
    id
  }
}
```


KirillDogadin-std:
@valiafetisov the boilerplate linked in the original issue is added. Since there was some rescoping done, please call the shot on wether i should implement the tests here as well.

src/modules/AuthPayload/model.ts
```diff
@@ -0,0 +1,9 @@
+import { objectType } from 'nexus/dist';
```
valiafetisov, line 1: I suggest to put all of those auth-related files into a single `module` called `Auth` or `User`

src/utils/auth.ts
```diff
@@ -0,0 +1,15 @@
...
+export function getUserId(authorization: string): string | null {
+  if (!authorization) {
+    return null;
+  }
+  const token = authorization.replace('Bearer ', '');
+  const verifiedToken = verify(token, SECRET) as { userId: string };
+  return verifiedToken && verifiedToken.userId;
+}
+
+export const APP_SECRET = JWT_SECRET;
```
valiafetisov, line 15: Please move this to the appropriate module

src/server.ts
```diff
@@ -0,0 +1,62 @@
...
+});
+
+const initializeApolloServer = (
+  httpServer: Server,
+  apollo: ApolloServer,
+  app: express.Application,
+  port: number,
+): (() => void) => {
+  apollo.applyMiddleware({ app });
+  serverCleanup = runSubscriptionServer(httpServer, apollo);
```
valiafetisov, line 38: Again, where is the subscription server requirement coming from? Why did you have to make any changes to the `createApolloServer`?

package.json
```diff
@@ -15,19 +15,32 @@
     "dependencies": {
         "@prisma/client": "^4.11.0",
         "altair-express-middleware": "^5.0.17",
+        "apollo-server-core": "^3.12.0",
         "apollo-server-express": "^3.12.0",
+        "bcrypt": "^5.1.0",
+        "bcryptjs": "^2.4.3",
```
valiafetisov, line 21: - Is it ok, that the package you're using [hasn't been updated for 6 years](https://www.npmjs.com/package/bcryptjs)?
- Why do you need two packages for the same thing if you're using only one?

src/utils/auth.ts
```diff
@@ -0,0 +1,15 @@
...
+import { JWT_SECRET } from '../env';
+
+const SECRET = JWT_SECRET;
+
+export function getUserId(authorization: string): string | null {
+  if (!authorization) {
+    return null;
+  }
+  const token = authorization.replace('Bearer ', '');
+  const verifiedToken = verify(token, SECRET) as { userId: string };
```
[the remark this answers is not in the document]
KirillDogadin-std: adjusted the error to be thrown on the stage of `shield`

package.json
KirillDogadin-std: ditched the `.js` one.

src/modules/User/resolvers.ts
```diff
@@ -0,0 +1,71 @@
...
+    };
+  },
+});
+
+export const signUp = mutationField('signUp', {
+  type: 'AuthPayload',
+  args: {
+    user: nonNull('UserCreateInput'),
+  },
+  resolve: async (_parent, { user }, ctx) => {
```
valiafetisov, line 53: Can you please introduce `AUTH_SIGNUP_ENABLED` in the same PR to control behaviour of this resolver?

src/modules/User/resolvers.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    const created = await ctx.prisma.user.create({
+      data: {
+        username,
+        password: hashedPassword,
+      },
+    });
+
+    return {
+      token: sign({ userId: created.id }, JWT_SECRET),
```
valiafetisov, line 67: - How long does this signature last?
- Does [ecosystem-api](https://github.com/makerdao-ses/ecosystem-api) jwt tokens expirable?
- How to invalidate this token, ie logout?
- How to update this token when it's expired?

src/server.ts
```diff
@@ -0,0 +1,59 @@
...
+import { createServer as createHttpServer } from 'http';
+import { ApolloServer } from 'apollo-server-express';
+import type express from 'express';
+import type { Disposable } from 'graphql-ws';
+import { ApolloServerPluginDrainHttpServer } from 'apollo-server-core';
+import { PORT, isDevelopment } from './env';
+import { createContext } from './context';
+import { schemaWithMiddleware } from './schema';
+
+let serverCleanup: Disposable;
```
valiafetisov, line 11: What is this for? Why did the `createApolloServer` and `startServer` had to change after being moved from `src/index.ts`?

src/modules/User/model.ts
```diff
@@ -0,0 +1,26 @@
...
+export const User = objectType({
+  name: 'User',
+  definition(t) {
+    t.string('id');
+    t.string('username');
+    t.string('password');
+  },
+});
+
+export const UserInputType = inputObjectType({
```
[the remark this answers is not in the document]
KirillDogadin-std: renamed and resused

src/modules/User/resolvers.ts
KirillDogadin-std: - added 7d expiration as in ecosystem api
- update: sign in again.
- logging out with tokens is not straightforward:
  - original code seems to not have this functionality
  - we would have to either decrease the expiry period from 7d to something like 10 min - 1h 
  - we could keep the list of expired tokens in the memory or db:
    - on each log out add the token to the global set
    - on each log out clean up the tokens that are past expiry
    - adjust the validation of the token with the check of it not being present in the expiry set.

any preference?

src/server.ts
KirillDogadin-std: this is actually a recommended plugin for server shut down
https://www.apollographql.com/docs/apollo-server/api/plugin/drain-http-server/

KirillDogadin-std: removed `server.ts` and simplified `index.ts`

src/index.ts
```diff
@@ -1,38 +1,50 @@
 import type { Server } from 'http';
 import { createServer as createHttpServer } from 'http';
 import { ApolloServer } from 'apollo-server-express';
-import { applyMiddleware } from 'graphql-middleware';
 import type express from 'express';
-import { PORT, isDevelopment } from './env';
+import type { Disposable } from 'graphql-ws';
```
valiafetisov, line 5: I believe this functionality is still related to the subscriptions

src/index.ts
```diff
@@ -1,38 +1,50 @@
...
+import { ApolloServerPluginDrainHttpServer } from 'apollo-server-core';
 import { createApp } from './app';
+import { PORT, isDevelopment } from './env';
 import { createContext } from './context';
-import { schema } from './schema';
+import { schemaWithMiddleware } from './schema';
 
-export const schemaWithMiddleware = applyMiddleware(schema);
+let serverCleanup: Disposable;
```
valiafetisov, line 12: You never actually assign anything to this variable

src/index.ts
```diff
@@ -1,38 +1,50 @@
...
+    ApolloServerPluginDrainHttpServer({ httpServer }),
+    {
+      async serverWillStart() {
+        return {
+          async drainServer() {
+            serverCleanup?.dispose();
+          },
+        };
+      },
+    },
```
valiafetisov, line 29:
```suggestion
```

src/modules/User/resolvers.ts
valiafetisov: I think the appropriate way to do logout functionality is by keeping the list of valid sessions instead of invalid ones. But I also see the advantage of storing only the logged out sessions since it's less comprisable. In any case, I would in principle not implement this logic ourselves, but instead look into authentication packages like [auth.js](https://authjs.dev/) to provide us with invalidation and renewal functionalities

To summarise,
- Let's not do it in this PR
- Before implementing it, let's investigate if there is appropriate package

src/permissions/index.ts
valiafetisov, line 1: Can this file be moved into the User/Auth module as well?

src/utils/auth.ts
valiafetisov, line 1: Can this file be moved into the User/Auth module as well?

src/modules/User/resolvers.ts
```diff
@@ -0,0 +1,70 @@
...
+import { compare, hash } from 'bcrypt';
+import { sign } from 'jsonwebtoken';
+import { AUTH_SIGNUP_ENABLED, JWT_EXPIRATION_PERIOD, JWT_SECRET } from '../../env';
+
+export const me = queryField('me', {
+  type: 'User',
+  resolve: (_, __, ctx) => {
+    if (!ctx.authVerificationResult || !('userId' in ctx.authVerificationResult)) {
+      throw new Error('Not authorized');
+    }
```
valiafetisov, line 11: Why do you need to separately check this and throw error inside the resolver if you have the whole `permissions` setup already? Because of the types? 
- If because of the types, can you please extend the types, so this check is not needed every time you access the `ctx.authVerificationResult` but the types are correctly propagated?
- If not because of the types, why there are need to be two places that check auth? Can we please use a single place inside `permissions`?

src/modules/User/resolvers.ts
```diff
@@ -0,0 +1,70 @@
...
+      throw new Error('user not found');
+    }
+    const passwordValid = (await compare(password, user.password || '')) || false;
+    if (!passwordValid) {
+      throw new Error('invalid password');
+    }
+    return {
+      token: sign({ userId: user.id }, JWT_SECRET, { expiresIn: JWT_EXPIRATION_PERIOD }),
+      user,
+    };
```
valiafetisov, line 42: One big advice we gave while looking into the [`ecosystem-api`](https://github.com/makerdao-ses/ecosystem-api) is to not spread the logic between model and the resolvers. Ideally, _all business logic_ should stay inside the model and the resolves should only define available queries/mutations and call a single method of the model. 

See second point of the [typeorm vs prisma discussion](https://github.com/makerdao-ses/switchboard-boilerplate/issues/2#issuecomment-1456504670) for the approaches on how to make this logic part of the model. To make it clear: I propose to replace all this code with `return User.signIn(userNamePass)` or `return ctx.prisma.user.signIn(userNamePass)`

(here and in all other resolvers)

src/index.ts
KirillDogadin-std: removed whatsovever.

src/utils/auth.ts
KirillDogadin-std: my gh does not show any code there. is this resolved? if not, pls recomment

image

src/permissions/index.ts
KirillDogadin-std: as in https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1135676006

src/modules/User/resolvers.ts
KirillDogadin-std: - it's not possible to adjust the context via `shield`
- the jwt validity object arrives to the shield and gets declined if invalid.
- then jwt object gets passed to the actual function. 
- the previous action was not taken into account by the typing system.

I can only come up with a hack where `userId` of the context gets hardcoded to be the string at all times. By default (where the id is not known from the jwt) it's something like 'unknown';

valiafetisov: I'm not sure I understand your answers, since you didn't actually answer to my questions.

> it's not possible to adjust the context via shield

I didn't ask for that, nor do I expect this is needed

> the jwt validity object arrives to the shield and gets declined if invalid

This I didn't get. What is the validity object? What means `declined`?

> the previous action was not taken into account by the typing system

By previous, do you mean "jwt object gets passed to the actual function". Then, my first suggestion stands true: please extend `ctx` type to have correct type for `ctx.authVerificationResult.userId` or just expose it under `ctx.userId` if that's simpler

> it's something like 'unknown'

This sounds terrible, see my suggestion above to fix types, instead of coming up with terrible workarounds

valiafetisov: Please answer my initial questions here https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1135625681

src/modules/User/resolvers.ts
```diff
@@ -0,0 +1,38 @@
...
+  },
+});
+
+export const signIn = mutationField('signIn', {
+  type: 'AuthPayload',
+  args: {
+    user: nonNull('UserNamePass'),
+  },
+  resolve: async (_parent, { user: userNamePass }, ctx) => {
+    const userInterface = new UserInterface(ctx.prisma.user);
```
valiafetisov, line 24: Please provide user interface inside context instead of recreating it every time. Probably a cleaner way would be to use `$extend` functionality [described here](https://www.prisma.io/docs/concepts/components/prisma-client/client-extensions/model)

src/modules/User/resolvers.ts
```diff
@@ -0,0 +1,38 @@
+import { queryField, mutationField, nonNull } from 'nexus/dist';
+import { UserInterface } from './model';
+
+export const me = queryField('me', {
+  type: 'User',
+  resolve: (_, __, ctx) => {
+    if (!ctx.authVerificationResult.userId) {
+      throw new Error('Unexpected middleware behaviour.');
+    }
```
valiafetisov, line 9: Please answer my initial questions https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1135912953 or in other way explain why this unnecessary check is needed? Why you don't need to check if `ctx.prisma.user` exists before accessing it?

src/utils/auth.ts
valiafetisov: It's a comment to a file, not code. I suggest to move the content of the file `src/utils/auth.ts` into the appropriate module. It's not just an util, it's a set of functions only usable in the context of authentication, right?

src/utils/auth.ts
```diff
@@ -0,0 +1,16 @@
...
+  }
+  return new ApolloError('Not authorized', 'NOT_AUTHORIZED');
+};
+
+export const ensureAuthenticated = (verification: JwtVerificationResult | null) => {
+  if (!verification || !('userId' in verification)) {
+    return getErrorFromJwtVerificationResult(verification);
+  }
+  return true;
+};
```
valiafetisov, line 16: https://github.com/makerdao-ses/switchboard-boilerplate/pull/16/files#r1135619883

src/context.ts
```diff
@@ -13,9 +19,30 @@ type CreateContextParams = {
...
+  if (!authorization) {
+    return { userId: null, error: 'Unhandled' };
+  }
+  const token = authorization.replace('Bearer ', '');
+  const verificationTokenResult = verify(token, JWT_SECRET, (err, decoded) => {
+    if (err) {
+      return { userId: null, error: err.name === 'TokenExpiredError' ? 'JwtExpired' : 'Unhandled' };
+    }
+    return decoded;
+  }) as unknown as JwtVerificationResult;
```
valiafetisov, line 32: Why do you have to retype it?

KirillDogadin-std: because the native typing seems to be simply incorrect. It claims to return void, although the actual return is the decoded object.

src/modules/User/resolvers.ts
KirillDogadin-std: > Why do you need to separately check this and throw error inside the resolver if you have the whole permissions setup already?

because of the types: ts cannot infer types because it does not see that the jwt verification has already happened.

So the following happens:
  1. data arrives to the endpoint
  2. the context is created: this implies calling the verification of the jwt token.
      - successful verification: user id is in the object `authVerificationResult`
      - unsuccessful verification: error identifier is in the object `authVerificationResult`
      - no token: null instead of the object
  3. If it's a shielded endpoint, the shielding logic activates and checks that the object formed on stage (2) is not indicating any errors
  4. endpoint logic is activated.

Types are generated by nexus, then in the respective `me` resolver ts expects the context as it has initially been defined. Regardless what happens on the stages (2) and (3). It's not a transfrom pipeline (`.validateToken().getMe()`) but a collection of building blocks for which typing has been generated (`makeApi ( tokenVerifier, getMeResolver )`) - so since each building block is not directly tied to another in the codebase (instead just accepts common context) , it cannot be known via inference what is the verification result.

I do not see a way where it's possible to make ts believe that the verification has happened and the user id is present for sure.

src/modules/User/resolvers.ts
KirillDogadin-std: https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1136824783

KirillDogadin-std: > Why you don't need to check if ctx.prisma.user exists before accessing it?

because it has been defined as the part of the context. in `context.ts`

valiafetisov (approved):
~~Implementation looks good~~, let's add tests here or in the next PR

Let's still resolve this conversation https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1138630281

src/modules/User/resolvers.ts
KirillDogadin-std: > Why can't you type the context to have correct union type to include all those types no matter if any of those building blocks executed or not?

this is (was) the case. and on successful verification it's not possible for ts to know if the id is present in the object or not because it expects the context as it is defined (your 3 bullet points).

> The other common way is to provide function into the context:

implemented this

src/context.ts
```diff
@@ -13,9 +18,30 @@ type CreateContextParams = {
...
+      throw new ApolloError(err.name === 'TokenExpiredError' ? 'Token expired' : 'Invalid authentication token');
+    }
+    return decoded;
+  }) as unknown as { userId: string };
+  return verificationTokenResult.userId;
+}
+
 export function createContext(params: CreateContextParams): Context {
+  const { req } = params;
+  const authorization = req.get('Authorization'); // for queries & mutations.
```
valiafetisov, line 37: Nit:
```suggestion
  const authorizationHeader = req.get('Authorization');
```

src/context.ts
```diff
@@ -13,9 +18,30 @@ type CreateContextParams = {
   connection?: unknown;
 };
 
+function getUserId(authorization?: string): string {
```
valiafetisov, line 21: Nit: I would accept cleaned up token here instead of the header

tests/helpers/server.ts
```diff
@@ -0,0 +1,43 @@
...
+
+interface TestContext {
+  client: GraphQLClient;
+}
+
+function getGraphqlTestContext() {
+  let serverInstance: Server | null = null;
+  return {
+    async before() {
+      const port = await getPort({ port: 3000 }); // 4
```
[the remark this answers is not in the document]
valiafetisov: Do you really need to specify hardcoded value instead of using `PORT` specified in the readme? Why using `get-port` package is necessary?

tests/auth.test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const signUpMutation = gql`
+  mutation {
+    signUp(user: { username: "asdf", password: "asdf" }) {
+      user {
+        username
+      }
+      token
+    }
+  }`;
```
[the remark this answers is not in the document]
KirillDogadin-std: nexus itself provides examples with hardcodng them
https://nexusjs.org/docs/getting-started/tutorial/chapter-testing-your-ap

i've tried to find something that generates it but did not discover any good looking packages that return a plain string with query there

src/modules/User/model.ts
```diff
@@ -0,0 +1,87 @@
...
+      created = await prisma.user.create({
+        data: {
+          username,
+          password: hashedPassword,
+        },
+      });
+    } catch (e) {
+      if (e instanceof PrismaClientKnownRequestError && e.code === 'P2002') {
+        throw new ApolloError('Username already taken', 'USERNAME_TAKEN');
+      }
```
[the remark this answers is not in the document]
KirillDogadin-std: already being tested in `Authentication: sign up with same username` , but there was a bug in there anyway.

tests/auth.test.ts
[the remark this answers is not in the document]
KirillDogadin-std: done

tests/auth.test.ts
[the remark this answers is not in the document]
valiafetisov: Oh, that's terrible. I wouldn't want to hardcode those things in plaintext, since it looses the point of having typescript and being able to validate that `signUp` even exist in the first place. Some kind of query builder will be a better choice here, I believe. Not sure about https://www.npmjs.com/package/gql-query-builder, maybe there is something better/provided by the nexus/nicely integrated with it?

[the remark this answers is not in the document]
valiafetisov: This looks promising https://github.com/remorses/genql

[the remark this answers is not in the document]
KirillDogadin-std: used this package. There's still no type check safety, but tests fail if the built query does not correspond to the defined structure. (e.g. if one fails to provide the correct type name for data sent with sign up mutation)

Did not find a better option of a builder.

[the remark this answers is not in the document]
KirillDogadin-std: i did not figure out how to handle the errors there without tests behaving incorrectly. Had to come up with a hack `catch(e) => e` in cases where an error is expected.

the behaviour is inconsistent and strange. If i am doing `expect.rejects` then e.g. test with incorrect password sign in fails because it does not reject. I wiggled around and also discovered that even if all the tests pass with `reject`, then there's a global error from vitest claiming that there's an unhandled error. Although i don't think there is one. And on top of all that the error that is thrown does not seem to correspond to a test case that the error points to.

generated/nexus.ts
```diff
@@ -1,199 +0,0 @@
-/**
- * This file was generated by Nexus Schema
- * Do not make changes to this file directly
- */
```
valiafetisov, line 4: Where did this file go?

generated/schema.graphql
```diff
@@ -1,6 +1,7 @@
-### This file was generated by Nexus Schema
-### Do not make changes to this file directly
```
valiafetisov, line 2: And this one is no longer generated?

package.json
```diff
@@ -15,26 +15,37 @@
...
         "nexus": "^1.3.0",
         "nexus-validation-plugin": "^0.2.0",
         "vite-node": "^0.29.2",
         "vitest": "^0.29.2"
     },
     "devDependencies": {
+        "@genql/cli": "^3.0.4",
+        "@types/bcrypt": "^5.0.0",
         "@types/express": "^4.17.17",
+        "@types/get-port": "^4.2.0",
```
valiafetisov, line 36: Do you still need this one?

tests/auth.test.ts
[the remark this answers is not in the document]
valiafetisov: > i did not figure out how to handle the errors

Do you describe the current state of the PR after you've integrated [genql](https://github.com/remorses/genql)? Do you suggest to revert back this integration? I see the potential problem in the fact that you use generated files [in the same code where you've generated them](https://github.com/makerdao-ses/switchboard-boilerplate/blob/6aa6420b4df0a36f1b0bd41cfd931f7056970b53/tests/helpers/server.ts#L9). But I'm also not a big fan of the amount of boilerplate code and strange packages (`undici`) that were added to integrate this, so I'm not against of reverting it back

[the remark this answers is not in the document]
KirillDogadin-std: > Do you describe the current state of the PR after you've integrated [genql](https://github.com/remorses/genql)?

yes

> Do you suggest to revert back this integration?

yes due to:
  - documentation is not so big and clear
  - the client error handling is not straightforward when combined with vitest
  - we would have to keep 2 generated folders around.

generated/schema.graphql
KirillDogadin-std: not relevant anymore due to https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1143081177 

initially i just yeeted the generation from new lib into the same folder in order to see if this would work and left it there because there was a topic in the linked thread i wanted to go over faster.

generated/nexus.ts
KirillDogadin-std: not relevant anymore due to https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1143081177 

initially i just yeeted the generation from new lib into the same folder in order to see if this would work and left it there because there was a topic in the linked thread i wanted to go over faster.

tests/auth.test.ts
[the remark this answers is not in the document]
valiafetisov: Alright. For the next issue, this also looks relevant and have more stars: [graphql-code-generator](https://github.com/dotansimha/graphql-code-generator)

tests/helpers/database.ts
```diff
@@ -0,0 +1,16 @@
+import { beforeEach, afterEach } from 'vitest';
+import { getPrisma } from '../../src/database';
+
+export function cleanDatabase() {
+  const clean = async () => {
+    const prisma = getPrisma();
+    await prisma.coreUnit.deleteMany();
+    await prisma.user.deleteMany();
```
valiafetisov, line 8: Nit: can you please generalise it in the next PR? Otherwise it's unclear that as soon as you create a new model, you also need to add line here. Eg https://github.com/prisma/docs/issues/451#issuecomment-1212842857

tests/helpers/database.ts
```diff
@@ -0,0 +1,16 @@
...
+  };
+  beforeEach(async () => {
+    await clean();
+  });
+  afterEach(async () => {
+    await clean();
+  });
```
valiafetisov, line 15: Nit: Do you actually need to do that before AND after. Let's remove unnecessary operations to speed up tests in the future

LukSteib (approved):
- was able to sign up users ✔️ 
Screenshot 2023-03-22 at 09 52 08
- querying `me` endpoint without auth header correctly set results in auth error ✔️

```

{
  "errors": [
    {
      "message": "Not authenticated",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "me"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR"

```

- querying `me` endpoint with auth header correctly set works ✔️

```
{
  "data": {
    "me": {
      "username": "lukas",
      "id": "e250741c-49a3-4c1d-87d9-94ce8a120db0"
    }
  }
}
```

Is there anything else that I'm supposed to check here @KirillDogadin-std?  
 

- signIn with wrong credentials not possible ✔️ 

```
{
  "errors": [
    {
      "message": "User not found",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "signIn"
      ],
      "extensions": {
        "code": "USER_NOT_FOUND",
```

KirillDogadin-std:
You can check also incorrect password submission, sign in without signing up.

src/modules/User/resolvers.ts
valiafetisov: Based on the [latest mock](https://github.com/makerdao-ses/switchboard-boilerplate/issues/24), I think we would need to create `session` table, where we would store session.id, expiration date, etc and than put this session.id into jwt, to later be able to mark some of those ids as `revoked` (on logout or manual revocation)

valiafetisov:
A short summary of the outstanding points for the next issues:
- Figure out type safe query builder, see https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1141792417
- Implement sessions, see https://github.com/makerdao-ses/switchboard-boilerplate/pull/16#discussion_r1135610008

Issue #17, opened 2023-03-14
Investigate logging libraries


LukSteib:
### Goal 
Research and decide on best logging library

### Context
We want to have an advanced logging library in place that allows to do error and performance logging. More input on original login requirement to inform the research: 
> should be able to toggle it, at least on a module level > to be plugged in to server logging infrastructure later

First library mentioned and that should be part of the investigation: [winston](https://github.com/winstonjs/winston)

### Tasks
- [x] Collect possible libraries
- [x] Propose qualities to make judgment
- [x] compare, discuss and conclude

KirillDogadin-std:
- https://tslog.js.org/#/
  - can be toggled: yes - https://tslog.js.org/#/?id=changing-settings-at-runtime
  - error logging: yes - https://tslog.js.org/#/?id=default-log-level
  - performance logging: yes - https://tslog.js.org/#/?id=default-log-level
  - other features: https://tslog.js.org/#/?id=all-features
    - typed
    - pretty & json output
    - Pretty Errors
    - Named Loggers
    - Sub-loggers with inheritance
    - prefixes
  - gh stars: 1k
  - last update: last week https://github.com/fullstack-build/tslog

- https://github.com/winstonjs/winston
  - can be toggled: yes - https://github.com/winstonjs/winston#using-logging-levels
  - error logging: yes - https://github.com/winstonjs/winston#logging-levels
  - performance logging: yes - https://github.com/winstonjs/winston#logging-levels
  - ts support: seems so https://github.com/winstonjs/winston/blob/master/test/typescript-definitions.ts
  - other features https://github.com/winstonjs/winston#table-of-contents: 
    - custom logging levels
    - handling uncaught exceptions
    - custom formats
    - filtering of info objects
    - log querying
    - gh stars 20k
    - last update 8 feb
- https://www.npmjs.com/package/typescript-logging
  - can be toggled: seems no - https://github.com/vauxite-org/typescript-logging not mentioned in the docs
  - error logging: yes (supports various log levels such as debug)
  - performance logging: yes (supports various log levels such as debug)
  - additional features:
    - not mentioned
  - gh stars: 97
  - last update: last week

before i proceed: i would appreciate to have a more specific requirement set (what we need from the logging) because in other case it seems like we could just choose one of the most popular frameworks and use it (winston / tslog) without problems with such narrow requirement set.

valiafetisov:
As specified in the issue description, main qualities are:
[...]

PR #21, opened 2023-03-16
Feat: add logging

### LINKED ISSUE #20: Implement logging library
### Goal
Logging library in place 

### Context
See investigaton #17 

### Tasks
- [ ] Implement logging library (based on the evaluation in research issue)

KirillDogadin-std:
closes #20

Tasks:

- [x] Install and configure it:
    - [x] Add appropriate info/debug/error/etc logs to the codebase
      - [x] To plain files
      - [x] To http requests
- [x] Forbid use of console via eslint rule
    - [ ] Can you maybe add a custom message to this error, suggesting use of a different library?
- [x] Add short example usage to the CONTRIBUTING readme (let's create one in the root!)
- [x] Make sure that in development it's nicely highlighted (ie via [pino-pretty](https://github.com/pinojs/pino-pretty))

___ 

- did not discover the way to overwrite eslint error message natively when violating the no-console rule
- looks like picrel
image


CONTRIBUTING.md
valiafetisov, line 1: I would also suggest to add usage commands:
- How to set custom logging level via cli (maybe you can also add `npm run debug` shortcut?)
- How to filter logging levels by prefix
- How to filter logs by specific folder (ie module)

src/modules/CoreUnit/resolvers.ts
```diff
@@ -2,14 +2,15 @@ import { stringArg, list, queryField } from 'nexus/dist';
...
 });
 
 export const coreUnit = queryField('coreUnit', {
   type: 'CoreUnit',
   args: { id: stringArg() },
-  resolve: (parent, { id }, ctx) => {
+  resolve: (_parent, { id }, ctx) => {
     if (!id) {
+      ctx.request.req.log.error('coreUnit resolver: id is undefined');
       throw new Error('please provide id');
```
valiafetisov, line 14: - Wouldn't `please provide id` already be in the logs?
- I would suggest to add a log to every resolver (or do they already produce logs as well?)

src/modules/CoreUnit/resolvers.ts
```diff
@@ -2,14 +2,15 @@ import { stringArg, list, queryField } from 'nexus/dist';
...
     }
     return ctx.prisma.coreUnit.findUnique({
```
valiafetisov, line 16: Let's add logging to the prisma as well. Can you check if an integration available for it like for express? Otherwise:
https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging

src/context.ts
```diff
@@ -1,19 +1,22 @@
...
 
 type CreateContextParams = {
-  req: express.Request;
+  req: express.Request & { log: pino.Logger };
   res: express.Response;
   connection?: unknown;
 };
 
 export function createContext(params: CreateContextParams): Context {
+  logger.debug('creating context with params: %o', params);
```
valiafetisov, line 19: Do we capitalise log messages or we don't? 🙂

valiafetisov (commented):
Please tick [the boxes](https://github.com/makerdao-ses/switchboard-boilerplate/issues/17#issuecomment-1471563047) that were addressed and comment on the reasons why some weren't 🙂

KirillDogadin-std:
> Please tick https://github.com/makerdao-ses/switchboard-boilerplate/issues/17#issuecomment-1471563047 that were addressed and comment on the reasons why some weren't

I can't edit them but will copy over here into the header

src/modules/CoreUnit/resolvers.ts
KirillDogadin-std: > Wouldn't please provide id already be in the logs?

The error that is thrown here gets captured and propagated to the api response.
Api response will contain all necessary info including traces (in case the env is `development`)

> I would suggest to add a log to every resolver (or do they already produce logs as well?)

there is only a single resolver apart from this one which directly uses prisma which in its turn already logs the information about the extraction. Should i still log the message with the response that is being sent out? i assume yes.

src/app.ts
```diff
@@ -2,10 +2,15 @@ import type { Express } from 'express';
 import cors from 'cors';
 import express from 'express';
 import expressPlayground from 'graphql-playground-middleware-express';
+import { expressLogger, getChildLogger, getModuleBinding } from './logger';
+
+const logger = getChildLogger({ msgPrefix: 'APP' }, { module: getModuleBinding(__filename) });
```
valiafetisov, line 7: If we keep pino, do you want to export a single function `getLogger({ prefix: 'OPTIONAL_PREFIX', module: __filename} )` instead of the two?

src/database.ts
```diff
@@ -1,15 +1,20 @@
...
 export const getPrisma = () => {
+  dbLogger.debug('Getting prisma client');
   if (!prisma) {
     prisma = new PrismaClient({
       datasources: {
         db: {
           url: 'file:./db.sqlite',
         },
       },
+      log: dbLogLevels,
```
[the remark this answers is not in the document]
KirillDogadin-std: i dont get this q.

Are you asking if pino prints it? or something else?

src/logger.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+  if (isPrefixFilterEmpty) {
+    return logger.child(bindings, localOptions);
+  }
+  const providedPrefix = localOptions.msgPrefix ?? '';
+  if (!formattedPrefixFilter.includes(providedPrefix)) {
+    localOptions.level = 'silent';
+  }
+  return logger.child(bindings, localOptions);
+};
```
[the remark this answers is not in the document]
KirillDogadin-std: there's some logic here for filtering that would also be necessary for winston filters. There you would have to write approximately the same amount of code. I've adjusted the current structure here to be more distinct.

src/context.ts
```diff
@@ -1,19 +1,24 @@
...
 import type express from 'express';
+import pino from 'pino';
 import { getPrisma } from './database';
+import { getModuleBinding, getChildLogger } from './logger';
+
+const logger = getChildLogger({ msgPrefix: 'CONTEXT' }, { module: getModuleBinding(__filename) });
 
 export interface Context {
-  request: { req: express.Request };
+  request: { req: express.Request & { log: pino.Logger } };
```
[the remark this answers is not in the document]
KirillDogadin-std: i don't get the question. 

on top it's variable, 
here it is the type defenition.

src/app.ts
KirillDogadin-std: refactored. no single function is imported and then called with minimum args

src/modules/CoreUnit/resolvers.ts
KirillDogadin-std: well, i've also updated the package of apollo from deprecated one.

valiafetisov (approved):

src/database.ts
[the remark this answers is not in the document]
valiafetisov: The log of the Prisma is not handled by pino, but it's rather just printed into the console. So if we to add a `transport` to pino, prisma logs will not end up there, right?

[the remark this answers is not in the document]
KirillDogadin-std: as per https://github.com/prisma/prisma/issues/10475 the logging with custom client is not yet implemented for prisma. Would you like me to extend the prisma client with custom logger as per the example in the same linked thread?

src/types.d.ts
```diff
@@ -0,0 +1,10 @@
+type LogLevel = 'trace' | 'debug' | 'info' | 'warn' | 'error' | 'fatal' | 'silent';
+type DbLogLevel = 'info' | 'query' | 'warn' | 'error';
```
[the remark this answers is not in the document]
KirillDogadin-std: Possible for pino, but prisma does not export its log level it seems.

src/types.d.ts
```diff
@@ -0,0 +1,12 @@
+import type { Level } from 'pino';
+
+type LogLevel = Level;
```
valiafetisov, line 3: Then, no need to redefine it here, right?

tests/helpers/server.ts
```diff
@@ -14,7 +14,7 @@ function getGraphqlTestContext() {
   return {
     async before() {
       const app = createApp();
-      serverInstance = await startServer(app, PORT);
+      serverInstance = await startServer(app);
       return new GraphQLClient(`http://0.0.0.0:${PORT}/graphql`);
```
valiafetisov, line 18: You no longer use provided `PORT` to start server, but still use it to connect? Why?

src/modules/User/model.ts
```diff
@@ -44,11 +44,11 @@ export function getUserCrud(prisma: PrismaClient) {
         },
       });
       if (!user) {
-        throw new ApolloError('User not found', 'USER_NOT_FOUND');
+        throw new GraphQLError('User not found', { extensions: { code: 'USER_NOT_FOUND' } });
```
valiafetisov, line 47: Curious, what's the difference between `ApolloError` and `GraphQLError` and why did you switch?

src/logger.ts
```diff
@@ -0,0 +1,86 @@
...
+const FILTERS = [filterModule, filterPrefix];
+
+const doesPassFilters = (config: {
+  options: pino.ChildLoggerOptions;
+  bindings: pino.Bindings;
+}): boolean => FILTERS.every((f) => f(config));
+
+export const expressLogger = pinoHttp({
+  level: httpLogLevel,
+  msgPrefix: '[EXPRESS]: ',
```
valiafetisov, line 53: Why does the `msgPrefix` here have a different format than all other ones, ie square brackets, space, etc?

tests/helpers/server.ts
KirillDogadin-std: adjusted to use the de-facto port.

src/modules/User/model.ts
KirillDogadin-std: Since i've updated the system from the deprecated https://www.npmjs.com/package/apollo-server-express to latest https://www.apollographql.com/docs/apollo-server/ , the error handling is now done according to the corresponding docs https://www.apollographql.com/docs/apollo-server/data/errors/#custom-errors

src/logger.ts
KirillDogadin-std: i did not call `formatPrefix` there but instead hardcoded the value.

src/database.ts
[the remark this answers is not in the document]
KirillDogadin-std: I've had to rewrite a bunch of `database.ts` lines:

  - i dont see the point in the getter: we can import prisma directly and there will be just a single instance of it anyways. I don't see the reason to import from `database` unless you want to access prisma, so building it first time in-place when it actually is required does not seem like a big win to me.
  - if we declare the prisma without instantiating it immediately then it derives and keeps its generics. Hence, all the event generics are already defined when we would do `prisma.$on`. With derived generics it forbids using `$on('query')` because type system thinks that we did not define logging event `'query'`. I.e. factual instance of prisma is more specific than the one we declare which blocks us from event handling.

src/logger.ts
valiafetisov, line 1: Nit: fine with me to have this file, if you know how to cover its logic with tests 🙂

valiafetisov (commented):
Please resolve merge conflicts!

src/logger.ts
KirillDogadin-std: added tests.

src/database.ts
```diff
@@ -1,24 +1,60 @@
...
-    });
-  }
-  const xprisma = prisma.$extends({
-    model: {
-      user: {
-        ...getUserCrud(prisma),
-      },
+const LOGGING_EVENTS: { level: Prisma.LogLevel; emit: 'event' }[] = [
+  {
+    emit: 'event',
```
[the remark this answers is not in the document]
KirillDogadin-std: https://www.prisma.io/docs/concepts/components/prisma-client/working-with-prismaclient/logging#log-to-stdout

this is done as in docs, 

this can be `emit` or `stdout`, but since we want to recapture the logging events, it's always `emit`.

I find this approach to be most simple and clear for understanding due to explicit-ness.

Added a comment with ref to documentation on top of this `const`

src/database.ts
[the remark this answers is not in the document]
valiafetisov: > we can import prisma directly and there will be just a single instance of it anyways

Can you please link a proof that no extra connection is opened when you import it every time. I think it's extra critical for the sqlite since it just doesn't work with multiply connections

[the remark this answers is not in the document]
KirillDogadin-std: https://www.prisma.io/docs/guides/performance-and-optimization/connection-management#prismaclient-in-long-running-applications

the official recommendation

https://stackoverflow.com/questions/46937494/es6-variable-import-by-reference-or-copy

stackoverflow answer with another reflink in it

src/env.ts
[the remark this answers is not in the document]
KirillDogadin-std: Via verbal discussion: 
  - removed "api execution ended" message because now we have functionality in place so starting a server runs this api forever.

Reason for the message to appear at the first place:
  - `startServer` started to return a value (server) instead of spawning server and returning void (this change was needed for the test context init) which resulted into triggering `then` immediately.

LukSteib (approved):
```
[18:09:22.807] DEBUG (23903): [APP] Creating app
    module: "src/app.ts"
[18:09:22.811] DEBUG (23903): [SERVER] Starting server
    module: "src/server.ts"
[18:09:22.880] INFO (23903): [SERVER] Running on 3000
    module: "src/server.ts"
```

Only judging by snippet above, I like the connotation. No further review done from my side on this one. Approved

PR #25, opened 2023-03-22
Improve test coverage

### LINKED ISSUE #23: Enforce high code coverage
### Goal
Ensure high code coverage

### Context
We're aiming for a high code coverage.
Target value for now: 90%

Let's ensure with respective tests

### Tasks
- [ ] Create tests to ensure code coverage of at least 90%

KirillDogadin-std:
close #23

- should i do other improvements in the same pr? (e.g. ones outstanding from auth test PR)?

- ci swears if coverage is low (as requested)
  - cause commit: [abc32ca](https://github.com/makerdao-ses/switchboard-boilerplate/pull/25/commits/abc32ca046255c6811eaafd37de969dc94aec046)
  - error can be seen at https://github.com/makerdao-ses/switchboard-boilerplate/pull/25/commits/f4aa1fc58101eb5149a2be0d4f98d88f12ff2d22 (different because linting was off before)
- there're some code pieces that make no sense to be tested via integration tests. Ignored these

### DIFF
--- a/src/env/getters.ts
+++ b/src/env/getters.ts
@@ -1,0 +1,8 @@
+export const getJwtSecret = (): string => {
+  if (!process.env.JWT_SECRET) {
+    if (process.env.NODE_ENV === 'production') {
+      throw new Error('JWT_SECRET is not defined');
+    }
+  }
+  return process.env.JWT_SECRET || 'dev';
+};
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -4,9 +4,9 @@ export default defineConfig({
   test: {
     coverage: {
       provider: 'istanbul',
-      lines: 0,
-      functions: 0,
-      statements: 0,
+      lines: 90,
+      functions: 90,
+      statements: 90,
       all: true,
     },
     singleThread: true,
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -1,10 +1,12 @@
 import { defineConfig, UserConfigExport } from 'vitest/config';
 
 export const getVitestConfig = (fullyCoveredModulePaths?: string[]): UserConfigExport => {
-  const coverage = !!fullyCoveredModulePaths && fullyCoveredModulePaths.length !== 0 ? 100 : 90;
+  const enableFullCoverage = !!fullyCoveredModulePaths && fullyCoveredModulePaths.length !== 0;
+  const coverage = enableFullCoverage ? 100 : 90;
   return {
     test: {
       coverage: {
+        enabled: !enableFullCoverage,
         provider: 'istanbul',
         lines: coverage,
         functions: coverage,
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -6,7 +6,7 @@ export const getVitestConfig = (fullyCoveredModulePaths?: string[]): UserConfigE
   return {
     test: {
       coverage: {
-        enabled: !enableFullCoverage,
+        enabled: true,
         provider: 'istanbul',
         lines: coverage,
         functions: coverage,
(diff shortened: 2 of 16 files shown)


valiafetisov:
Do I miss it or there is no overall code coverage percentage in the coverage report? https://github.com/makerdao-ses/switchboard-boilerplate/actions/runs/4490008456/jobs/7896539944
- Therefore, what's the current number? Is it possible to add it to the report?
- Can we top 90 up and add some other generic tests?
- Can you please validate that it's enforced by submitting a commit (and reverting it later one) which break the CI coverage check?
- Is there a way to enforce higher code coverage for the modules folder? I.e. 100%?

KirillDogadin-std:
> Can you please validate that it's enforced by submitting a commit (and reverting it later one) which break the CI coverage check?

as mentioned in the top comment, this is done. Here's the direct link to ci https://github.com/makerdao-ses/switchboard-boilerplate/actions/runs/4489993993/jobs/7896508154

src/env/getters.ts
```diff
@@ -0,0 +1,8 @@
+export const getJwtSecret = (): string => {
+  if (!process.env.JWT_SECRET) {
+    if (process.env.NODE_ENV === 'production') {
+      throw new Error('JWT_SECRET is not defined');
+    }
+  }
+  return process.env.JWT_SECRET || 'dev';
+};
```
KirillDogadin-std, line 8: i wanted to test this so refactored `env.ts`

valiafetisov:
>> Is there a way to enforce higher code coverage for the modules folder? I.e. 100%?
>
> i did not find a ready-to-go solution for this

Reading istanbuljs readme I found [mentioning of how to run multiply coverage reports](https://github.com/istanbuljs/nyc#combining-reports-from-multiple-runs) per project. I think this approach fulfils the requirement: we can just have two `nyc` configs to set two different glob patterns + coverage numbers. Linked [`c8`](https://github.com/bcoe/c8) can even receive those parameters in as cli arguments, so no extra configs files are needed. Worst case: two vitest configs, one extending the other and being executed two times

KirillDogadin-std:
So this is obviously a hack. Apart from running the tests twice:

 - the config would essentially be duped, and watched after in order to keep it as similar as possible.
 - the `defineConfig` accepts the object with a lot of fields with subfields that might be defined or not.
   - so to dedupe it i put this in a getter because alternative of defining an object (of config type) and then modifying its copy gives you a type error (because you are not allowed to modify optional fields before checking for their presence). Let me know if you see the issue with potential future extendability of this approach.
- In the mid term running tests twice will cost a lot of time because `appropriateTime * 2 == aLotOfTime` is evaluated into `true`
  - which then will force to optimize and also define tests that are run in each of the suits. Which dictates test module structure.

vitest.config.ts
```diff
@@ -1,14 +1,21 @@
...
-export default defineConfig({
-  test: {
-    coverage: {
-      provider: 'istanbul',
-      lines: 0,
-      functions: 0,
-      statements: 0,
-      all: true,
+export const getVitestConfig = (fullyCoveredModulePaths?: string[]): UserConfigExport => {
+  const enableFullCoverage = !!fullyCoveredModulePaths && fullyCoveredModulePaths.length !== 0;
```
valiafetisov, line 4: Nit: Not sure why you've decided to complicate this and split the logic into two files instead of just exporting the pure config object here and overwriting its parts in the other file

Otherwise, you can also read the cli flag here and have a single config file

valiafetisov (approved):

LukSteib (approved):

PR #28, opened 2023-03-27
Implement sessions

### LINKED ISSUE #26: Implement sessions
### Goal

User can list created sessions and revoke any of them including current one

### Context

We've implemented auth in #16, but there is no logout/revoke functionality yet. In order to implement that, the suggestion is to:
- Create a table called `Session` with 
  - `id`, `createdAt`, `createdBy`, `name`
  - `referenceExpiryDate` – token expiration date stored just for the UI (the expiration should be validated via the jwt itself)
  - `referenceTokenId` – token start and end (eg `34a3...0c97`)
  - `revokedAt` – empty or a date when it was revoked
- Store `sessionId` inside jwt token (instead of the userId or in addition to?)
- Rename `getUserId` to `getUser`, use `sessionId` to check if the session is still valid, otherwise throw
- Create endpoint to list your own sessions
- Create endpoint to revoke your own token by session id
- Create endpoint to create new tokens with user-specified `name` and `expiryDurationSeconds`
- Come up with the way to distinguish user-created and sign in/sign up-created sessions (eg empty `name`)

### Tasks
- [ ] Ask questions about implementation and context, clarify implementation proposal
- [ ] Implement proposal

KirillDogadin-std:
close #26 

how to manually test:

1. sign up / sign in, .. (https://github.com/makerdao-ses/switchboard-boilerplate/pull/16)
```gql
mutation {
  signUp(user: { username: "asdf", password: "asdf" }) {
    token
  }
}
```
2. create session

```gql
mutation {
  createSession(
    session: {name: "2"}
  )
  { session {referenceExpiryDate} }
}
```
```gql
mutation {
  createSession(
    session: {name: "12", expiryDurationSeconds: 3600}
  )
  { session {referenceExpiryDate, name} }
}
```
```gql
mutation {
  createSession(
    session: {name: "12", expiryDurationSeconds: 3600}
  )
  { session {referenceExpiryDate, name} }
}
```
3. list sessions
```gql
query {
  sessions {referenceExpiryDate, name, isUserCreated, id} // more fields here possible, see the docs in the playground
}
```
4. revoke session
```gql
mutation {
  revokeSession(sessionId: "ID HERE") {
    id
  }
}

### DIFF
--- a/api/tests/helpers/database.ts
+++ b/api/tests/helpers/database.ts
@@ -1,31 +1,14 @@
 import { beforeEach, afterEach } from 'vitest';
 import prisma from '../../src/database';
 
-let tableNameList: string[] | undefined;
-
-function formatTableNameToPrismaName(tableName: string) {
-  return tableName.charAt(0).toLowerCase() + tableName.slice(1);
-}
-
-const getTableNames = async () => {
-  if (typeof tableNameList !== 'undefined') {
-    return tableNameList;
-  }
-  const tableNames = (
-    await prisma.$queryRaw<{ name: string }[]>`SELECT name FROM sqlite_schema WHERE type ='table' AND name NOT LIKE 'sqlite_%';`
-  );
-  tableNameList = tableNames.map((table) => formatTableNameToPrismaName(table.name));
-  return tableNameList;
-};
-
 export function cleanDatabase() {
   const clean = async () => {
-    const tablenames = await getTableNames();
-    await tablenames.slice(1).reduce(async (acc, tableName) => {
-      const table: any = prisma[tableName as any];
-      return acc.then(() => table.deleteMany());
-    }, (prisma[tablenames[0] as any] as any).deleteMany() as Promise<void>);
-    // can't type properly tablenames received via raw query.
+    // TODO: when migrate to postgres,
+    // replace below with
+    // https://www.prisma.io/docs/concepts/components/prisma-client/crud#deleting-all-data-with-raw-sql--truncate
+    await prisma.session.deleteMany();
+    await prisma.user.deleteMany();
+    await prisma.coreUnit.deleteMany();
   };
   beforeEach(async () => {
     await clean();

### CODE REVIEW

**Comment by KirillDogadin-std** [comment]
just noticed: will have to extend revoke to handle case where session does not belong to the caller

File: src/helpers/token.ts [hunk 1]
```diff
@@ -0,0 +1,12 @@
+import ms from 'ms';
+import { sign } from 'jsonwebtoken';
+import { JWT_SECRET, JWT_EXPIRATION_PERIOD } from '../env';
+
+export const format = (token: string) => `${token.slice(0, 3)}...${token.slice(-3)}`;
+
+export const generate = (sessionId: string, expireDate?: Date) => {
+  const expiresIn = expireDate
+    ? ms(expireDate.getTime() - Date.now())
+    : JWT_EXPIRATION_PERIOD;
```
valiafetisov, line 10: To be more precise, I would do the opposite: 
- I would accept duration in originally supplied format here
- Return jwt token
- Provide another function which parses jwt token and returns `exp` (encoded into it) as a js Date
- Write Date parsed from the token into the database

src/modules/Session/model.ts
```diff
@@ -0,0 +1,108 @@
...
+    t.nonNull.boolean('isUserCreated');
+    t.string('name');
+    t.date('revokedAt');
+  },
+});
+
+export const SessionCreate = inputObjectType({
+  name: 'SessionCreate',
+  definition(t) {
+    t.date('referenceExpiryDate');
```
valiafetisov, line 24: The original suggestion was to accept number of seconds, otherwise you need to validate past date, figure out timezone difference between frontend and server, etc

Also, the idea behind `reference*` prefix is to signal that the value stored in the database is not actually used to validate token (so editing it will not make tokens work longer or shorter). Using this prefix for `input` is confusing

src/modules/Session/model.ts
```diff
@@ -0,0 +1,108 @@
...
+  const createdSession = await newSession(prisma, createData);
+  return {
+    token: createdToken,
+    session: createdSession,
+  };
+}
+
+export function getSessionCrud(prisma: PrismaClient) {
+  return {
+    listSessions: async (userId: string) => listSessions(prisma, userId),
```
valiafetisov, line 101: Instead of redefining those functions and passing them prisma every time, please closely follow the pattern that was agreed and approved in `getUserCrud`: one function that returns object with all methods and shares prisma context

src/modules/Session/resolvers.ts
```diff
@@ -0,0 +1,45 @@
...
+    _parent,
+    { session }: { session: { referenceExpiryDate?: Date; name: string } },
+    ctx,
+  ) => {
+    const id = await ctx.getUserId();
+    const {
+      session: createdSession,
+      token,
+    } = await ctx.prisma.session.generateTokenAndSession(id, session);
+    return { session: createdSession, token };
```
valiafetisov, line 43: Nit: can't you directly return it?
```suggestion
    return await ctx.prisma.session.generateTokenAndSession(id, session);
```

src/modules/Session/model.ts
```diff
@@ -0,0 +1,108 @@
...
+      data: {
+        revokedAt: new Date(),
+      },
+    });
+  } catch (e) {
+    throw new ApolloError('Failed to update session', 'SESSION_UPDATE_FAILED');
+  }
+}
+
+export async function generateTokenAndSession(
```
valiafetisov, line 72: If it's exported as part of the session model (available via prisma.session.generateTokenAndSession) let's not export it again and access it separately. Same applies to other functions in every model. There is just one export and all those functions should be only accessible through prisma

src/context.ts
```diff
@@ -27,8 +30,20 @@ function getUserId(token?: string): string {
       throw new ApolloError(err.name === 'TokenExpiredError' ? 'Token expired' : 'Invalid authentication token');
     }
     return decoded;
-  }) as unknown as { userId: string };
-  return verificationTokenResult.userId;
+  }) as unknown as { sessionId: string };
```
valiafetisov, line 33: Do you also want to move `verify` to the newly created `token.ts` under `validateAndParse` and properly type it there?

valiafetisov: Also, I thought `ApolloError`s were replaced with graphql errors in a previous PR, but I still can see one above. Please take care of it

src/context.ts
```diff
@@ -27,8 +30,20 @@ function getUserId(token?: string): string {
...
+      id: sessionId,
+    },
+    include: {
+      creator: true,
+    },
+  });
+  if (session.revokedAt && session.revokedAt < new Date()) {
+    throw new ApolloError('Session expired');
+  }
+  return session.creator.id;
```
valiafetisov, line 46: Let's return user object and rename `getUserId` to `getUser` for clarity (as suggested in the issue description)

src/context.ts
```diff
@@ -27,8 +30,20 @@ function getUserId(token?: string): string {
...
+    where: {
+      id: sessionId,
+    },
+    include: {
+      creator: true,
+    },
+  });
+  if (session.revokedAt && session.revokedAt < new Date()) {
+    throw new ApolloError('Session expired');
+  }
```
valiafetisov, line 45: I believe this whole part should be part of the session model, eg `prisma.session.getUser(token)`

src/context.ts
```diff
@@ -18,7 +21,7 @@ type CreateContextParams = {
   connection?: unknown;
 };
 
-function getUserId(token?: string): string {
+async function getUserId(xprisma: XPrismaClient, token?: string): Promise<string> {
```
valiafetisov, line 24: Why do you need to pass prisma client here?

src/env/getters.ts
```diff
@@ -6,3 +8,15 @@ export const getJwtSecret = (): string => {
...
+export const getJwtExpirationPeriod = (): string => {
+  if (!process.env.JWT_EXPIRATION_PERIOD_SECONDS) {
+    return '7d';
+  }
+  const expirationSeconds = Number(process.env.JWT_EXPIRATION_PERIOD_SECONDS);
+  if (Number.isNaN(expirationSeconds)) {
+    throw new Error('JWT_EXPIRATION_PERIOD_SECONDS must be a number');
+  }
+  // https://www.npmjs.com/package/jsonwebtoken for `expiresIn` format
+  return ms(expirationSeconds * 1000);
```
valiafetisov, line 21: The documentation specifies that it accept plain seconds OR `ms` format, why did you decide to introduce `ms` in this PR?

src/context.ts
valiafetisov: Also, I don't think you need to check `session.revokedAt < new Date()` since there is no functionality to revoke session in the future that you provide

src/helpers/token.ts
KirillDogadin-std: Replaced with number of seconds. 

The last 2 bullet points dont make sense to me, why would i write data about expiry after parsing the token if i already have the data about expiry before encoding the token? would it not be easier to map the seconds into expiry date directly without passing it through JWT encode-decode?

src/env/getters.ts
KirillDogadin-std: 1. expiration period is provided through env variable:
    - it is imo nicer to provide a string over a long seconds number
2. `jsonwebtoken` docs says that the numeric value represents number of seconds (true)
3. the packages (`ms`) that `jsonwebtoken` references claim that numeric value represents the number of milliseconds (also true, so jsonwebtoken probably does *1000 under the hood)

so for (1) - human readability and (2) + (3) inconsistency, i would prefer to pass the expiration to the jwt encoder in a string format for explicitness.

alternatively it's obviously possible to just force everything to be seconds. and accept the downside of having to manually convert + accept absence of the explicit time units.

KirillDogadin-std: 1. it is imo nicer to provide a string over a long seconds number for readability and debugging purposes
2. `jsonwebtoken` docs says that the numeric value represents number of seconds (true)
3. the packages (`ms`) that `jsonwebtoken` references claim that numeric value represents the number of milliseconds (also true, so jsonwebtoken probably does *1000 under the hood)

so for (1) - human readability and (2) + (3) inconsistency, i would prefer to pass the expiration to the jwt encoder in a string format for explicitness.

alternatively it's obviously possible to just force everything to be seconds. and accept the downside of having to manually convert + accept absence of the explicit time units.

src/modules/Session/model.ts
KirillDogadin-std: each resolver for user has now 1 more line , but i don't think it's critical since it looks more like a middleware pipe then a set of particular actions.

src/modules/Session/model.ts
KirillDogadin-std: fixed as part of https://github.com/makerdao-ses/switchboard-boilerplate/pull/28#discussion_r1151494134

src/helpers/token.ts
valiafetisov: > why would i write data about expiry after parsing the token if i already have the data about expiry before encoding the token?

Well, since you don't control what's written into the JWT, because you don't how they process supplied period and turn it into unix timestamp, how they handle timezones, etc. It can be, that you provide one thing, but something else written to the token (eg you provide no expiration date, but something else is written there), so to be factual you should decode it as suggested

api/src/modules/Session/resolvers.ts
```diff
@@ -0,0 +1,41 @@
...
+});
+
+export const create = mutationField('createSession', {
+  type: 'SessionCreateOutput',
+  args: {
+    session: nonNull('SessionCreate'),
+  },
+  resolve: async (
+    _parent,
+    { session }: { session: { referenceExpiryDate?: Date; name: string } },
```
valiafetisov, line 35: Why does the `{ session }` is not properly typed according to `SessionCreate`? Does it mean the input is not validated?

valiafetisov: [According to this comment](https://github.com/makerdao-ses/switchboard-boilerplate/pull/28#discussion_r1151487407), please:

- [ ] Don't call whatever you receive here `referenceExpiryDate`
- [ ] Call it `expiresInSeconds`
- [ ] Receive `number`

api/src/modules/Session/resolvers.ts
```diff
@@ -0,0 +1,41 @@
...
+    ctx,
+  ) => {
+    const { id } = await ctx.getUser();
+    return ctx.prisma.session.generateTokenAndSession(id, session, true);
```
valiafetisov, line 39:
```suggestion
    const { id } = await ctx.getUser();
    return ctx.prisma.session.generateApiSession(id, session);
```

src/env/getters.ts
valiafetisov: > it is imo nicer to provide a string over a long seconds number for readability and debugging purposes

Fine with me, you can accept strings from the env variable if you are able to validate them with `ms`

api/src/modules/User/resolvers.ts
```diff
@@ -17,13 +20,25 @@ export const signIn = mutationField('signIn', {
   args: {
     user: nonNull('UserNamePass'),
   },
-  resolve: async (_parent, { user: userNamePass }, ctx) => ctx.prisma.user.signIn(userNamePass),
+  resolve: async (_parent, { user: userNamePass }, ctx) => {
+    const user = await ctx.prisma.user.validatePassword(userNamePass);
+    return ctx.prisma.session.generateTokenAndSession(
+      user.id,
+      { expiryDurationSeconds: ms(JWT_EXPIRATION_PERIOD) / 1000, name: 'Sign in' },
+    );
```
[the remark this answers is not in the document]
KirillDogadin-std: went with the second option.

api/src/modules/Session/resolvers.ts
KirillDogadin-std: `{ session: { referenceExpiryDate?: Date; name: string } }` for some reason did not produce a warning with the mismatch (i.e. did not suggest to do `unknown as ...`). Just removing this one already checks the boxes.

api/src/modules/Session/resolvers.ts
KirillDogadin-std: named it `createCustomSession` to match `createSignInSession` pattern of name.

src/env/getters.ts
valiafetisov: > it is imo nicer to provide a string over a long seconds number for readability and debugging purposes

> for the env i first validate the env var to be a number.

Aren't those contradictory statements? 

I suggest you define `JWT_EXPIRATION_PERIOD` as
```
const JWT_EXPIRATION_PERIOD = process.env.JWT_EXPIRATION_PERIOD ?? '7d'
```
and then validate `JWT_EXPIRATION_PERIOD` no matter if it's declared in the code or supplied via env

api/tests/helpers/database.ts
```diff
@@ -5,6 +5,7 @@ export function cleanDatabase() {
   const clean = async () => {
     await prisma.coreUnit.deleteMany();
     await prisma.user.deleteMany();
+    await prisma.session.deleteMany();
```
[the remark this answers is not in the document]
KirillDogadin-std: i've tried these both **briefly** before submitting this line.

- both break tests.
  - with strange errors
  - not a surprise for direct query since it's for psql

- the github solutions (the linked one and those below) rely on specific prisma versions apparently (prisma obj layout) and on top of this there's a good point at https://github.com/prisma/docs/issues/451#issuecomment-1354500062

Given that suggested programmatic solutions do not work out cleanly, i would ask the following:

Should i write raw sqlite process here or should i just write manual command that deletes and recreates the db file? latter might be less efficient in runtime, but is simpler. If we intend to migrate to psql eventually, it's a quick and ez way forward.

api/src/modules/Session/model.ts
```diff
@@ -0,0 +1,142 @@
...
+      { expiryDurationSeconds: ms(JWT_EXPIRATION_PERIOD) / 1000, name: 'Sign up' },
+    ),
+    createCustomSession: async (
+      userId: string,
+      session: { expiryDurationSeconds?: number | null; name: string },
+      isUserCreated: boolean = false,
+    ) => generateTokenAndSession(prisma, userId, session, isUserCreated),
+    async getUser(
+      token?: string,
+    ) {
```
valiafetisov, line 117: I suggested two options:
- `user.getUserByToken(token)`
- `session.getSessionByToken(token)`

And you made `session.getUser()` instead without bringing any arguments. Can you please elaborate why did you need to mix session with user?
```suggestion
    async getSessionByToken(
      token?: string,
    ) {
```

api/tests/helpers/database.ts
```diff
@@ -1,10 +1,30 @@
...
+
 export function cleanDatabase() {
   const clean = async () => {
-    await prisma.coreUnit.deleteMany();
-    await prisma.user.deleteMany();
+    const tablenames = await getTableNames();
+    await tablenames.slice(1).reduce(async (acc, tableName) => {
+      const table: any = prisma[tableName as any];
+      return acc.then(() => table.deleteMany());
+    }, (prisma[tablenames[0] as any] as any).deleteMany() as Promise<void>); // can't type properly tablenames received via raw query.
```
valiafetisov, line 27: Why did the [official solution](https://www.prisma.io/docs/concepts/components/prisma-client/crud#deleting-all-data-with-raw-sql--truncate) not work? Can you please bring more context to help you solve it? This code looks like a mess due to:
- multiply raw queries with hardcoded things like `sqlite_` – what happens when we move to postgres?
- constant type re-mapping
- unexplained `slice(1)`
- global variables

If none of the proposed clean versions work, and I can not help you to figure them out, I would prefer to go back to manual deletion (because it cleaner and more readable) than to approve this code

api/src/modules/Session/resolvers.ts
valiafetisov: Can you please answer to the two original questions, because they might help me understand what happened here?
- Why does the { session } is not properly typed according to SessionCreate?
- Does it mean the input is not validated?

api/prisma/schema.prisma
```diff
@@ -23,4 +23,20 @@ model User {
...
+  createdAt           DateTime  @default(now())
+  createdBy           String
+  referenceExpiryDate DateTime?
+  name                String?
+  revokedAt           DateTime?
+  referenceTokenId    String
+  isUserCreated       Boolean   @default(false)
+
+  creator User @relation(fields: [createdBy], references: [id], onDelete: Cascade)
+  @@unique([createdBy, id])
```
valiafetisov, line 41: Not sure why do you specify it like that, since you treat `session.id` as unique property. Let's make `id` primary/unique instead

api/src/modules/User/model.ts
```diff
@@ -30,13 +27,13 @@ export const AuthPayload = objectType({
...
     t.string('token');
-    t.field('user', { type: 'User' });
+    t.field('session', { type: 'Session' });
   },
 });
 
 export function getUserCrud(prisma: PrismaClient) {
   return {
-    signIn: async (userNamePass: { username: string; password: string }) => {
+    validatePassword: async (userNamePass: { username: string; password: string }) => {
```
valiafetisov, line 36: Nit: let's rename it to `getUserByUsernamePassword`, cause `validate` sounds like it doesn't return anything

api/src/modules/Utility/model.ts
valiafetisov, line 1: Nit: I don't think there should be a Utility module. Please move it to something like `graphql/*` subfolder where apollo server can also live together with context, etc

src/env/getters.ts
KirillDogadin-std: > Aren't those contradictory statements?

bad wording perhaps. "nicer when a programm provides you a ...". Since it's almost 0 effort to convert once when you're setting the env but ~`almost-0`*`logLength` effort when you're reading the logs where these are seconds.

> How is it known?

due to the logic present in `env/getters.ts` which rejects non-numbers.

____

adjusted the logic now. It's possible to give both `12` and `7h` which will be correct. On incorrect values it will throw.

api/src/modules/Session/resolvers.ts
valiafetisov: > Should i deep dive into nexus docs for specific reason / look for TS rule so that this does not happen?

My question was, to reiterate, why it was needed to retype those types in the first place? How did `{ session }: { session: { referenceExpiryDate?: Date; name: string } }` happened in the first place? What type `session` have inside the resolver? `any` or is it properly typed according to the nexus input?

KirillDogadin-std: > My question was, to reiterate, why it was needed to retype those types in the first place? 

this was probably an automatic instinct to type the variable for explicitness. There was no strong reason behind it for sure

> What type session have inside the resolver? any or is it properly typed according to the nexus input?

see the image - seems properly typed according to the nexus.

Image might look confusing since the popup is not directly visible. will add code snippet with typing hint below the image

image

```
(parameter) session: {
    expiryDurationSeconds?: number | null | undefined;
    name: string;
}
```

api/src/graphql/context.ts
```diff
@@ -0,0 +1,38 @@
...
+  logger.trace('Creating context with params: %o', params);
+  const { req } = params;
+  const authorizationHeader = req.get('Authorization');
+  const token = authorizationHeader?.replace('Bearer ', '');
+
+  return {
+    request: params,
+    prisma,
+    apolloLogger,
+    getSessionByToken: async () => prisma.session.getSessionByToken(token),
```
valiafetisov, line 36:
```suggestion
    getSession: async () => prisma.session.getSessionByToken(token),
```
You don't need to provide token to the function in the context, so the name is misleading

api/src/modules/Session/resolvers.ts
valiafetisov: Thanks!

valiafetisov (approved):

LukSteib (approved):

LukSteib:
I know already merged but in retrospect one thing catched my eye:

```
{
        "referenceExpiryDate": "2023-03-30T15:47:08.000Z",
        "name": "2",
        "isUserCreated": true,
        "id": "8ef9fca1-74ea-47e3-9d65-16ed9065e22b",
        "revokedAt": "2023-03-30T15:02:24.848Z"
      },
```
- I've created the session without specifying duration at 16:47 CET
- I've revoked the session at 17:02 CET
-> Is it possible that the two dates from the snippet above are calculated differently or calculated at different places (client vs. server) with a mismatch in computation?

PR #39, opened 2023-04-03
Replace nexus with pothos

KirillDogadin-std:


api/src/modules/Session/crud.ts
KirillDogadin-std, line 1: @valiafetisov 

advice needed

This file violates the agreed before structure (i.e. crud is in separate from `model` file)

if these functionalities (calls on `builder` and factory function for crud) are located at the same module, then importing the crud from the `models.ts` into `database.ts` causes an error. See example of it in collapsable below. If these are split, no error is thrown.
<details><summary> error snippet </summary> 

```
file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/@pothos/plugin-prisma/esm/util/get-client.js:15
    return (_client__baseDmmf = client._baseDmmf) !== null && _client__baseDmmf !== void 0 ? _client__baseDmmf : client._dmmf;
                                       ^

TypeError: Cannot read properties of undefined (reading '_baseDmmf')
    at getDMMF (file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/@pothos/plugin-prisma/src/util/get-client.ts:64:17)
    at SchemaBuilder.prismaObject (file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/@pothos/plugin-prisma/src/schema-builder.ts:34:35)
    at /Users/kirilldogadin/work/switchboard-boilerplate/api/src/modules/Session/model.ts:3:32
    at ViteNodeRunner.directRequest (file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/vite-node/dist/client.mjs:315:5)
    at ViteNodeRunner.cachedRequest (file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/vite-node/dist/client.mjs:159:14)
    at ViteNodeRunner.dependencyRequest (file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/vite-node/dist/client.mjs:207:14)
    at /Users/kirilldogadin/work/switchboard-boilerplate/api/src/modules/User/model.ts:3:31
    at ViteNodeRunner.directRequest (file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/vite-node/dist/client.mjs:315:5)
    at ViteNodeRunner.cachedRequest (file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/vite-node/dist/client.mjs:159:14)
    at ViteNodeRunner.dependencyRequest (file:///Users/kirilldogadin/work/switchboard-boilerplate/api/node_modules/vite-node/dist/client.mjs:207:14)
```
</details>

<details><summary>The logic of building schema is:</summary>

  - the logic of `builder` object creation is split among multiple files. Each of the calls produces a corresponding type that can be reused in different `builder` call. E.g.

```ts
// file1.ts
const builder = ...
// file2.ts
export const Field = builder.prismaField( ... );
// file3.ts
builder.queryField(
  ...
  Field
  ...
)
// file4.ts
builder.toSchema()
```
  - so to receive the final `builder` from which the schema is generated, one has to execute all the files. I.e. import them all somewhere
```ts
import 'modelsA'
import 'modelsB'
import 'resolversA'
import 'resolversB'

export const schema = builder.toSchema()
```
 - in the pseudo code snippet above (prev bullet point) resolver or model files can also import the exports from each other.
</details>

<details><summary>Resolution that i would imagine should work:</summary>

it would be handy if one could execute the building process via a function call instead of `import`ing. But this to my understanding prevents the dev from directly exporting the generated type from the module and requries one to pass it around between functions. Which is nasty imo.

</details>

<details><summary> Possible resolution that i might have overseen </summary>

At some point there was a line that did `import {getUserCrud} from 'models'` instead of `import 'models'`. This caused an error. I suspect that i might oversee some fine details about how imports work under the hood and i might just not use the proper import statements. When `import {getUserCrud} from 'models``ing i would expect that all the file is executed and on top of this the function is awailable in the module that imports as per [this](https://stackoverflow.com/questions/37325667/does-es6-module-importing-execute-the-code-inside-the-imported-file)

</details>

api/tests/auxilary.test.ts
```diff
@@ -40,10 +40,10 @@ test('Env: jwt expiration in seconds format', async () => {
 });
 
 describe('Healthz', () => {
-  vi.mock('../src/database');
-
   test('healthz: returns 200', async () => {
-    prisma.user.findFirst.mockResolvedValueOnce({ id: '1', username: 'asdf', password: 'asdf' });
```
KirillDogadin-std, line 46: this was mocked before. Mock removed because pothos pluging expects a prisma object as part of it config. Hence, in case prisma gets mocked, it breaks the pothos plugin and everything crashes.

> (I like gql stack.)

api/tests/auth.test.ts
```diff
@@ -1,120 +1,99 @@
...
+  const expiry = new Date(signInResponse.signIn.session.referenceExpiryDate!).getTime();
   const now = new Date().getTime();
 
   const token = signInResponse?.signIn?.token;
-  ctx.client.setHeader('Authorization', `Bearer ${token}`);
+  setHeader({ Authorization: `Bearer ${token}` });
 
   // wait until token expires
   // eslint-disable-next-line no-promise-executor-return
   await new Promise((resolve) => setTimeout(resolve, expiry - now));
```
KirillDogadin-std, line 90: this breaks on my local machine every other (every second) test run when the whole suite runs.

"Breaks" means: test times out because this promise waits too long. It waits too long because the `expiry` variable has a high value (although the env var value is mocked) for no visible reason

> (i like vitest)

KirillDogadin-std (commented):
above comments indicate the current problems that require resolution

valiafetisov:
Currently, while running `npm run test` locally, I receive a lot of:

```
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: listen EADDRINUSE: address already in use :::3000
 ❯ Server.setupListenHandle [as _listen2] node:net:1740:16
 ❯ listenInCluster node:net:1788:12
 ❯ Server.listen node:net:1876:7
 ❯ Module.startServer src/graphql/server.ts:490:21
     48|     logger.info(`Running on ${PORT}`);
     49|   });
     50| };
       |   ^
     51| 
 ❯ Object.before tests/helpers/server.ts:14:24
 ❯ tests/helpers/server.ts:33:5
 ❯ callSuiteHook node_modules/@vitest/runner/dist/index.js:373:23
 ❯ runTest node_modules/@vitest/runner/dist/index.js:436:28
 ❯ runSuite node_modules/@vitest/runner/dist/index.js:534:15

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: {
  "address": "::",
  "code": "EADDRINUSE",
  "errno": -48,
  "port": 3000,
  "syscall": "listen",
}
This error originated in "tests/auth.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
The latest test that might've caused the error is "Authentication: access protected endpoint without valid token". It might mean one of the following:
- The error was thrown, while Vitest was running this test.
- This was the last recorded test before the error was thrown, if error originated after test finished its execution.
```

What am I doing wrong or how am I suppose to run tests locally?

valiafetisov:
I mean, nothing is running there. Moreover, if I run `PORT=3000 npm run test` it throws the same error (but obviously, with the different port number). Does vitest run tests in parallel which results in multiply servers starting on the same port?

KirillDogadin-std:
i've run this locally on my machine and it works + the ci agrees with me (it runs the tests successfully)

> Does vitest run tests in parallel which results in multiply servers starting on the same port?

no, the vitest config covers this.

___

stupid assumption, but miiight work. 
can you delete node_modules and reinstall?
I am not sure but i might have had the same issue and it was gone after reisntall

valiafetisov:
I think that `Authentication: token expiration error` is actually failing for the same reason as https://github.com/makerdao-ses/switchboard-boilerplate/pull/28#issuecomment-1490484089: looking at the additional logs, I can see that [`referenceExpiryDate` returned](https://github.com/makerdao-ses/switchboard-boilerplate/blob/7a0657d79755d72338d2d019649269ca4041c18f/api/tests/auth.test.ts#L88) is 7 days in the future and expectedly can not be awaited:

Screenshot 2023-04-11 at 15 32 08

Moving forward I would:
- Create expirable token using graphql API instead mocking `JWT_EXPIRATION_PERIOD`
- Rewrite [this line](https://github.com/makerdao-ses/switchboard-boilerplate/blob/7a0657d79755d72338d2d019649269ca4041c18f/api/tests/auth.test.ts#L96) to _always_ await 1 second so that the test fail due to the token not being expired instead of the timeout
- If this doesn't solve the problem, investigate and fix https://github.com/makerdao-ses/switchboard-boilerplate/pull/28#issuecomment-1490484089 in the same PR

KirillDogadin-std:
https://github.com/makerdao-ses/switchboard-boilerplate/pull/39#issuecomment-1504872204

i've adjusted this test, although the main problem still remains: there's something wrong with how the spy mocks work. There're 2 more tests that might fail: healthz and signup disabled test. Both work on with mocks right now. As mentioned verbally, i can remove the mocks of the env vars and manage this via direct overwrites. With healthz i can ignore the 500 case for now.

valiafetisov:
Regarding the second issue with failing `healthz: returns 500`, here is [the docs](https://vitest.dev/guide/mocking.html) (in the original formatting):

> However spies can only help you **spy** on functions, they are not able to alter the implementation of those functions. In the case where we do need to create a fake (or mocked) version of a function we can use `vi.fn()` ([read more here](https://vitest.dev/api/vi.html#vi-fn)).

Looking at the code, I can see you're using `spyOn` to mock returned value 🙂 

https://github.com/makerdao-ses/switchboard-boilerplate/blob/7b0bf6d06112f6eb7311e6fa79109c43131dca0a/api/tests/auxilary.test.ts#L49

api/tests/auth.test.ts
```diff
@@ -81,19 +81,31 @@ test('Authentication: access protected endpoint without valid token', async () =
...
+        referenceExpiryDate: true,
+      },
+      __args: {
+        session: {
+          expiryDurationSeconds: 1,
+          name: 'test',
+        },
+      },
+    },
+  });
```
valiafetisov, line 102: As far as I understand, this is not correct use of the graphql client (hinted by the use of internal `__args` variable). Instead, you should first build the mutation in one place (eg `const createSession = ctx.client.mutation({...})`) and then use mutation in your test `await createSession({ name: 'test', expiryDurationSeconds: 1)` otherwise you need to repeat a lot of unnecessary parts of the code. But I can't find the documentation of `client` you're using since it's completely generated and uses `fetch` under the hood. Can you please point to the documentation based on which this mutation is written? Or can you link to the docs on the lib that generates the client?

KirillDogadin-std:
> Regarding the second issue with failing healthz: returns 500, here is [the docs](https://vitest.dev/guide/mocking.html) (in the original formatting):

we're reading the same docs in different ways.

if you follow the link you've added and scroll to the example [here](https://vitest.dev/guide/mocking.html#example-1) you can find the following happening:

```
const spy = vi.spyOn(messages, 'getLatest')
...
spy.mockImplementationOnce(() => 'access-restricted')
```

which is equivalent to my current approach:

```
 vi.spyOn(prisma.user, 'findFirst').mockImplementationOnce(async () => null as any); 
```

are you sure that with this context my approach is still invalid according to the docs?
  - if yes: why? and why is it allowed by ts to `mockImplementation` of the object member? is vitest typed incorrectly?

valiafetisov:
> and scroll to the example [here](https://vitest.dev/guide/mocking.html#example-1) you can find the following happening

I also seen this example, but then the [quoted sentence](https://github.com/makerdao-ses/switchboard-boilerplate/pull/39#issuecomment-1504970444) doesn't make any sense.

The most disappointing is that `npx vitest run -t "healthz: returns"` always works without glitching. Only running it together with other tests results in error. 

I also tried to mock it via `vi.mock('./database', () => {})` but couldn't make it work quickly. My main suspicion now is that prisma does some magic under the hood with the prisma client which makes it incompatible with tests. They actually have a special guide on how to mock it with jest. But at this point I suggest to drop this particular test and create an issue to enable it in the future.

api/tests/auth.test.ts
```diff
@@ -1,120 +1,117 @@
...
+  const meResponse: GenqlError = (await getMe(ctx.client).catch((e) => e));
+  expect(meResponse.errors[0].message).toBe('Token expired');
 });
 
 test('Authentication: sign up disabled', async () => {
   vi.spyOn(env, 'AUTH_SIGNUP_ENABLED', 'get').mockReturnValue(false);
-  const response = (await executeGraphQlQuery(getSignUpMutation())) as any;
+  const response: GenqlError = await signUp(ctx.client).catch((e) => e);
   expect(response.errors[0].message).toBe('Sign up is disabled');
 });
```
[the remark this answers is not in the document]
valiafetisov: Let's remove AUTH_SIGNUP_ENABLED and make it always enabled (:

KirillDogadin-std:
Oki, all the comments addressed, nasty tests removed.

But there's apparently more: coverage does not see that the line are covered with the tests. Consider this post to be a status update :)

valiafetisov:
As discussed in the daily: there is another problem with weird `vitest` behaviour regarding coverage that is also glitching. Based on this, let's deprioritize this PR and focus on more important functionality for now.

KirillDogadin-std:
Okai,

verbally discussed:

this turns out to be a fight against the tool and not the problem. Already sunk too much time into bug hunt. We let this go for now and stick to the current gql library

Issue #22, opened 2023-03-16
Integrate documentation into frontend


LukSteib:
### Goal 
Research and decide on best doc standardisation library

### Context
Initial requirement:
> Setup something like [tsdoc](https://tsdoc.org/) (can be enforced via a linter rule)

### Tasks
- [ ] Collect possible alternatives
- [ ] Propose qualities to make judgment
- [ ] compare, discuss and conclude

valiafetisov:
### Requirements
In general, there are at least two different problems
1. Provide self-explainable interface for querying data, for example:
    - `1A` Graphql queries that have description visible in the graphql explorer
    - `1B` Graphql explorer that have something like a swagger: an interface to execute known queries with the ability to add descriptions to them within the code
2. Provide self-explainable codebase for other developers who would want to extend it
    - `2A` Enforce docstrings for all functions
    - `2B` Build static documentation of the code

Ideally, I imagine that those those two things can be combined (ie via an extension), but if that's not possible, only the `2` should be done in this PR. In any case, the final outcome should be:
- [ ] [eslint] rules to enforce docstrings in our codebase
  - [ ] Comprehensive docscrings for all existing functions
- [ ] A command to build [single-page] html-based documentation of the code
- [ ] An integration into our existing playground/server setup
  - [ ] Served as a separate static page by the existing express server or extended/interlinked with the playground

### Research tasks
- [ ] Research solutions for 1 and 2 separately
- [ ] Identify if 1 and 2 can be merged together in some way
- [ ] Research `tsdoc` alternatives, ie [jsdoc seem to support typescript](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html) and is way more established
- [ ] Research available build templates, eg https://ankdev.me/clean-jsdoc-theme/v4/
  - [ ] Are there single-page templates or can any existing template be configured to be a single page or how the pagination works? Does it follow code structure?
- [ ] Research on how to integrate proposed solutions with [`graphql-playground-middleware-express`](https://www.npmjs.com/package/graphql-playground-middleware-express) or express itself
- [ ] Propose a concrete way forward

valiafetisov:
During today's standup we've though if maybe [docus](https://github.com/nuxt-themes/docus) can fulfil some of the outlined requirements?

KirillDogadin-std:
1. markdown
  With nuxt 3 we're able to utilize the `content` directory (plugin) that is able to provide access to markdown files:
[...]

PR #38, opened 2023-04-03
Frontend auth

valiafetisov:
This PR implements login + token listing, generation and revoking functionality

Current status:

Screenshot 2023-04-06 at 18 05 20
Screenshot 2023-04-06 at 18 04 52

PR #42, opened 2023-04-06
Documentation integration

KirillDogadin-std:
Closes https://github.com/makerdao-ses/switchboard-boilerplate/issues/22

Screenshot 2023-04-14 at 13 17 52


KirillDogadin-std:
[d246620](https://github.com/makerdao-ses/switchboard-boilerplate/pull/42/commits/d2466203cf357ce273dfa82596b47941909a1b61)

is ugly proof of concept for the documentation generation approach.
There's a bunch of useless files (it's a draft at this point) there. 

For the sake of documenting how this state was achieved:

1. Grab docs in md format,
2. put them into `content/docs` directory.
3. run the following command

```
 npx concat-md content/docs --file-name-as-title --dir-name-as-title > JOJO.md  
```
4. copy the generated file (JOJO.md) into the `content/docs/JOJO.md` and run the frontend
5. open `localhost:3000/docs/jojo`
6. ???
7. seems to be not half-bad of a result.

frontend/pages/docs.vue
```diff
@@ -0,0 +1,45 @@
...
+            </div>
+          </div>
+          <div class="w-3/4 sm:max-lg:w-full p-4">
+            <ContentRenderer :value="doc" class="border-black" />
+          </div>
+        </div>
+      </template>
+    </ContentDoc>
+  </div>
+</template>
```
KirillDogadin-std, line 39: this seems suboptimal. Tried using naive ui's layout component, but it breaks `#` links.

valiafetisov: What exactly do you find suboptimal?

frontend/README.md
```diff
@@ -17,3 +17,20 @@ You can build application for production using `npm run build` and then locally
...
+This process goes as follows:
+
+1. Put the documentation into the `/docs` directory
+2. Run the script for generation of the single documentation file that frontend consumes.
+
+```bash
+$ npm run mkdocs
+```
+
+3. The documentation that is generated will end up at `./content/documentation/index.md` file.
```
valiafetisov, line 36: Since it's not the final process, I suggest to make it very short, describing current and intended flow with TODO

Regarding intended flow: I expect that we will need to get docs from the `../api/node_modules` and therefore dockerfile will need to be edited to include bigger context (probably can be done in the next PR)

frontend/app.vue
```diff
@@ -11,4 +11,8 @@
 html, body {
   @apply bg-neutral-50;
 }
+
+.Body {
+    @apply h-[calc(100vh-88px)];
+}
```
valiafetisov, line 17: This is a terrible hack (to hardcode height of the header in css). I suggest to remove it (here and in the other place) and instead use combination of `flex-grow` and `h-full`. Overall, you can maybe wait for the #38 merged cause it also fixes some of the related issues (eg fixed header)

frontend/package.json
```diff
@@ -16,18 +16,27 @@
     "postinstall": "nuxt prepare",
     "lint": "eslint --max-warnings 0 .",
     "typecheck": "nuxt typecheck",
-    "start": "NODE_ENV=production node ."
+    "start": "NODE_ENV=production node .",
+    "mkdocs": "vite-node ./scripts/generateDocs.ts"
```
valiafetisov, line 20:
```suggestion
    "generateDocs": "vite-node ./scripts/generateDocs.ts"
```

frontend/components/doc/TableOfContents.vue
```diff
@@ -0,0 +1,42 @@
+<script setup lang="ts">
+const props = defineProps<{
+  tocLinks: Array<{
+    id: string;
+    text: string;
+    children: { id: string; text: string }[];
```
valiafetisov, line 6: Are there only two levels of links? Can't it be recursive instead?

frontend/scripts/generateDocs.ts
```diff
@@ -0,0 +1,25 @@
+import fs from 'fs'
+import path from 'path'
+import { concatMdSync } from 'concat-md'
+import { mkdirpSync } from 'mkdirp'
```
valiafetisov, line 4: I think `fs` it powerful enough to support that:

```
fs.mkdirSync(targetDir, { recursive: true });
```

nginx.conf
```diff
@@ -7,8 +7,8 @@ http {
             proxy_pass http://frontend:3000;
         }
         # api
-        location /api/ {
-            rewrite /api/(.*) /$1 break;
+        location /backend/ {
+            rewrite /backend/(.*) /$1 break;
```
valiafetisov, line 11: If you change it in nginx, let's also change it in the ingress (I think you will need to wait for Denis or merge fork yourself into this repo first)

frontend/package.json
```diff
@@ -16,18 +16,27 @@
...
+    "concat-md": "^0.5.1",
     "eslint": "^8.33.0",
+    "merge-md": "^1.0.3",
+    "mkdirp": "^2.1.6",
     "nuxt": "^3.2.3",
     "typescript": "^4.9.5",
     "vue-tsc": "^1.0.24"
+  },
+  "dependencies": {
+    "@vueuse/core": "^9.13.0",
```
valiafetisov, line 39: Where do you use `vueuse`?

frontend/pages/docs.vue
KirillDogadin-std: i would imagine that there's supposed to be a way to do the same with components instead of pure html. 

This one seems cluttered. On top of this while playing around with tailwind i've encountered some cases where the css classes break the hash url access as well. 

If the current state of html only seems fine to you, i could roll with it as well.

frontend/components/doc/TableOfContents.vue
KirillDogadin-std: - i beleive it should not be recursive: 
  1. less straightforward implementation
  2. unpredictable behaviour can happen due to recursion
  3. https://content.nuxtjs.org/api/configuration#toc defines the default depth of the object that i'm working with here which is 2, which i wrote the component for
  
it seems reasonable to adjust this on-the-go in separate issue and redefine toc object + component when needed.

nginx.conf
```diff
@@ -7,8 +7,8 @@ http {
             proxy_pass http://frontend:3000;
         }
         # api
-        location /api/ {
-            rewrite /api/(.*) /$1 break;
+        location /backend/ {
```
KirillDogadin-std, line 10: changed the name of the `/api` location to not overlap with the one `/api` path of nuxt3.

kubernetes/main.switchboard.k8s.sidestream.tech/values-api.yml
```diff
@@ -54,7 +54,7 @@ ingress:
     nginx.ingress.kubernetes.io/rewrite-target: /$1
   hosts:
   - url: main.switchboard.k8s.sidestream.tech
-    path: /api/(.*)
+    path: /backend/(.*)
```
KirillDogadin-std, line 57: adjusted api value path to correspond with the new nginx config

KirillDogadin-std (commented):
@DeusAvalon pls validate these changes (i.e. are these sound or will things break because i missed some other config)

nginx.conf
KirillDogadin-std: https://github.com/makerdao-ses/switchboard-boilerplate/pull/42#pullrequestreview-1383084858

addressed according to the link above

frontend/package.json
```diff
@@ -16,22 +16,27 @@
     "postinstall": "nuxt prepare",
     "lint": "eslint --max-warnings 0 .",
     "typecheck": "nuxt typecheck",
-    "start": "NODE_ENV=production node ."
+    "start": "NODE_ENV=production node .",
+    "generateDocs": "vite-node ./scripts/generateDocs.ts"
   },
   "devDependencies": {
-    "@huntersofbook/naive-ui-nuxt": "^0.5.1",
```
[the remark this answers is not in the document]
valiafetisov: I've downgraded it since it stopped properly working with tailwind: more info [in another project of ours](https://github.com/sidestream-tech/t4f-gxmanager/pull/52). So the suggestion is to keep it down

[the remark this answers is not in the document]
valiafetisov: Fixed via 809fc9a

valiafetisov:
Currently, I'm experiencing weird behaviour which can be described as "when frontend service is started, api service is no longer process requests". To reproduce:
1. Start api service via `cd api && PORT=4000 npm run dev`
2. Check that api service works by visiting http://localhost:4000
3. Start frontend service in another tab via `cd frontend && API_ORIGIN=http://localhost:4000 npm run dev`
4. Check that api service no longer return anything by visiting http://localhost:4000

KirillDogadin-std:
i have done something wrong probably, here's problem that i have after step 2:

image

valiafetisov:
> i have done something wrong probably

Nope, you see the UI, which already indicates that backend works. If you want to test if it can actually make graphql requests, you need to edit `http://localhost:4000/api/graphql` to be `http://localhost:4000/graphql` in the UI

KirillDogadin-std:
> Can you check if you can reproduce it (also, after wiping node_modules)?

wiped node modules, was able to reproduce. I get infinite loading on localhost:4000

valiafetisov:
To my knowledge it did not exist on my branch #38 (but maybe I didn't wipe `node_modules` while development?) so I'm wandering what happened here

KirillDogadin-std:
i did not pinpoint the reason right away.
But it seems to be not a problem if i run on port `4001` instead of `4000`.

valiafetisov (approved):

PR #46, opened 2023-04-13
Add origin restriction to session token

### LINKED ISSUE #41: Implement optional token restriction by origin
### Goal
User can restrict the token by origin

### Context
During UI conceptualization and thoughts on auth introduced the requirement for optional origin restriction of tokens.
The mock below shows how we'll implement origin restriction possibility into the UI. 
The user will be able to specify origin URL and all requests coming with this token but having another origin will be blocked

### Assets
image

### Tasks
- [x] Implement functionality to restrict tokens by origin
   - [x] default value: no origin restriction at all
- [x] ensure that requests of token that is restricted but coming from different origin fail

KirillDogadin-std:
close #41 
image


api/src/graphql/context.ts
```diff
@@ -28,11 +28,12 @@ export function createContext(params: CreateContextParams): Context {
   const { req } = params;
   const authorizationHeader = req.get('Authorization');
   const token = authorizationHeader?.replace('Bearer ', '');
+  const origin = req.get('Origin');
```
KirillDogadin-std, line 31: @valiafetisov don't have prior xp on this one, googling did not yield me a definite answer.

Is relying on the origin header safe? can't it be forged? e.g. in tests i manage to put any origin i want and hence i imagine that this approach is not safe.

valiafetisov: The header can not be changed by the JavaScript running in the browser, but can be set in many other environments (like curl). It is sufficient for the issue

api/src/helpers/origin.ts
```diff
@@ -0,0 +1,36 @@
+import { GraphQLError } from 'graphql';
+import wildcard from 'wildcard-match';
+
+export function isOriginValid(originParam: string): boolean {
+  if (originParam === '*') {
+    return true;
+  }
+  if (originParam.includes(' ')) {
+    return false;
+  }
```
valiafetisov, line 10: I believe it would be better for the user to trim every provided origin automatically instead of just invalidating it

api/src/helpers/origin.ts
```diff
@@ -0,0 +1,36 @@
...
+  const origins = originParam.split(',');
+  return origins.some((origin) => {
+    if (!origin.startsWith('http://') && !origin.startsWith('https://')) {
+      return false;
```
valiafetisov, line 14: Instead of returning true or false, validators should normally return error message that let the user understand what's wrong (with enough information to fix it)

api/src/helpers/origin.ts
valiafetisov, line 1: Why is this file becoming a general helper while in fact it's a part of the session/user module?

api/src/helpers/origin.ts
```diff
@@ -0,0 +1,36 @@
...
+    }
+    return true;
+  });
+}
+export function throwGQLErrorIfOriginDisallowed(
+  originRestriction: string,
+  originReceived?: string,
+) {
+  if (originRestriction !== '*') {
```
valiafetisov, line 23: Please prefer to return earlier
```suggestion
  if (originRestriction === '*') {
      return;
  }
```

api/src/modules/Session/model.ts
```diff
@@ -115,19 +124,20 @@ export function getSessionCrud(prisma: PrismaClient) {
     createSignInSession: async (userId: string) => generateTokenAndSession(
       prisma,
       userId,
-      { expiryDurationSeconds: ms(JWT_EXPIRATION_PERIOD) / 1000, name: 'Sign in' },
+      { expiryDurationSeconds: ms(JWT_EXPIRATION_PERIOD) / 1000, name: 'Sign in', originRestriction: '*' },
```
valiafetisov, line 127: I think sign up and sign in tokens should be limited to the origin they've sent from. For example, sign up done via our ui should display `https://main.switchboard.k8s.sidestream.tech`

api/src/graphql/context.ts
valiafetisov: Related question to @DeusAvalon. Is it enough to just `req.get('Origin')` on the application level to get the actual origin (not `frontend:3000`) or should ingress be additionally configured to pass proper header to the application? Or does it come via something like `x-origin`?

api/prisma/schema.prisma
```diff
@@ -36,6 +36,7 @@ model Session {
   revokedAt           DateTime?
   referenceTokenId    String
   isUserCreated       Boolean   @default(false)
+  originRestriction   String
```
valiafetisov, line 39: `originRestriction = '*'` does sounds to me as "all origins are restricted", therefore, the suggestion is to rename it to `allowedOrigins`

api/src/graphql/context.ts
DeusAvalon: Im pretty sure `req.get('Origin')` on Application level should give the correct Origin (actual Ingress URL) as the Ingress does not strip/replace any headers in our current configuration.

api/src/modules/Session/model.ts
KirillDogadin-std: added the env var for it, readme row and kube values adjustment

valiafetisov: > added the env var for it

As specified, "limited to the origin they've _sent from_": ie, a frontend making the sign in request will automatically provide `https://main.switchboard.k8s.sidestream.tech` inside origin, without any additional env variables required. If the origin is not set in the request, you can fallback to `*`

api/src/graphql/context.ts
valiafetisov: Hmm, then what header contains internal `frontend:3000`? And what is [`proxy-add-original-uri-header`](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#proxy-add-original-uri-header) for?

If possible, I would ask kirill to enable staging deployments on this branch to validate that it indeed works, but as the repo wasn't transferred yet, we would need to test (and potentially fix) it after this PR is merged, via our fork

api/src/modules/Session/model.ts
```diff
@@ -36,6 +41,40 @@ export const SessionCreateOutput = objectType({
...
 });
 
+// Actions
+
+function validateOrigin(originParam: string): void {
+  if (originParam === '*') {
+    return;
+  }
+  const trimmedOriginParam = originParam.trim();
+  const origins = trimmedOriginParam.split(',');
```
valiafetisov, line 51: This will also trim spaces between origins

```suggestion
  const trimmedOriginParam = originParam.trim();
  const origins = trimmedOriginParam.split(',').map(origin => origin.trim());
```

But since you trim it here, I would call it `parseOriginMarkup` and return cleanedup value which is then stored in the database

api/src/modules/Session/model.ts
```diff
@@ -36,6 +41,40 @@ export const SessionCreateOutput = objectType({
   },
 });
 
+// Actions
```
valiafetisov, line 44: Feel free to put it into the `Session/helpers.ts` file, to not expend this file over the limit

frontend/components/auth/SessionTable.vue
```diff
@@ -67,6 +71,13 @@ const columns = [
       return 'never'
     }
   },
+  {
+    title: 'Origin restriction',
```
valiafetisov, line 75:
```suggestion
    title: 'Allowed origins',
```

valiafetisov (dismissed):

api/src/modules/Session/model.ts
KirillDogadin-std: sorry for missing this one.

valiafetisov:
Last question: any specific reason why [wildcard-match](https://www.npmjs.com/package/wildcard-match) package was used over [matcher](https://www.npmjs.com/package/matcher). They are both wasn't updated for 2 years, but `matcher` is way [more popular](https://npmtrends.com/matcher-vs-wildcard-vs-wildcard-match). I also wonder if there are even better alternatives (or if those packages small enough to require no updates)?

valiafetisov (approved):

api/src/graphql/context.ts
DeusAvalon: Hmm i see, yeah im not 100% sure either i would need to actually deploy a service and check which headers are being parsed to the service.

But the staging deployment could actually be done by me manually deploying it based on this branch. This way we can pre-validate if its actually working. If this would work for you i would organize this tmrw with @KirillDogadin-std together

KirillDogadin-std:
I don't think there's enough info for me to get what went wrong.
Feel free to dial me up for a screenshare.
Here's what worked for me:

- single frontend (port 3000)
- single api (port 4001)
- create token with restriction to `http://localhost:5555`
- throw `me` request using this token in `authorization` header.
- receive origin restriction error.

KirillDogadin-std:
likely this is the reason:

  - we serve the playground through iframe
  - actual location of the playground is the api
  - therefore request's origin is that of api
  - in my case i run the api on port 4001
  - and the origin header sent via the playground is
image

does not matter how many frontends you start

i suggest creating the following functional test:

- run 1 frontend
- run 1 api
- create 2 tokens. 1st being `http://localhost:*` 2nd being `https://google.com*`
- use both tokens for `me` request, 1st success , 2nd failure

KirillDogadin-std:
Dockerized run yielded:

1. Run dockercompose up
2. add the token with origin restriction to localhost
3. add the token with origin restricion to google
4. run `me` query in the playground with both tokens.
5. (2) runs, (3) does not run with appropriate error.

LukSteib (approved):

PR #49, opened 2023-04-14
Preserve auth header for playground

### LINKED ISSUE #45: Share auth with graphql playground
### Goal
User logged in to the frontend, doesn't need to execute `signUp` query or set proper headers to run queries

### Context
Since auth frontend was implemented in https://github.com/makerdao-ses/switchboard-boilerplate/pull/38, we want to improve UX of the playground users and do not require them to run additional queries there. Technically that would (I imagine) also check for token in cookies IF authorization header is not provided and add the cookie after the signIn/signUp is executed.

### Tasks
- [ ] Propose alternative solution if known
- [ ] Implement
- [ ] Write tests
- [ ] Ensure it works both with auth reader and with the cookie

KirillDogadin-std:
close #45

### DIFF
--- a/api/src/app.ts
+++ b/api/src/app.ts
@@ -1,6 +1,7 @@
 import type { Express } from 'express';
 import express from 'express';
 import expressPlayground from 'graphql-playground-middleware-express';
+import cookierParser from 'cookie-parser';
 import { getChildLogger } from './logger';
 import prisma from './database';
 
@@ -31,12 +32,18 @@ export const createApp = (): Express => {
 
   app.get(
     '/',
-    expressPlayground({
-      endpoint: '/api/graphql',
-      settings: {
-        'editor.theme': 'light',
-      },
-    }),
+    cookierParser(undefined, { decode: (value: string) => value }),
+    (req, res, next) => {
+      const globalHeaders = req.cookies['gql:default'];
+      const playgroundMiddleware = expressPlayground({
+        endpoint: '/api/graphql',
+        settings: {
+          'editor.theme': 'light',
+          'request.globalHeaders': { Authorization: `Bearer ${globalHeaders}` },
+        },
+      });
+      playgroundMiddleware(req, res, next);
+    },
   );
 
   return app;
--- a/api/package-lock.json
+++ b/api/package-lock.json
@@ -11,8 +11,10 @@
             "dependencies": {
                 "@apollo/server": "^4.5.0",
                 "@prisma/client": "^4.11.0",
+                "@types/cookie-parser": "^1.4.3",
                 "bcrypt": "^5.1.0",
                 "body-parser": "^1.20.2",
+                "cookie-parser": "^1.4.6",
                 "cors": "^2.8.5",
                 "dotenv": "^16.0.3",
                 "express": "^4.18.2",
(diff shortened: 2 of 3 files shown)


api/src/app.ts
```diff
@@ -31,12 +32,18 @@ export const createApp = (): Express => {
...
-      },
-    }),
+    cookierParser(undefined, { decode: (value: string) => value }),
+    (req, res, next) => {
+      const globalHeaders = req.cookies['gql:default'];
+      const playgroundMiddleware = expressPlayground({
+        endpoint: '/api/graphql',
+        settings: {
+          'editor.theme': 'light',
+          'request.globalHeaders': { Authorization: `Bearer ${globalHeaders}` },
```
valiafetisov, line 42: This will always set the reader, which does not fulfil the requirement defined in the issue `check for token in cookies IF authorization header is not provided`

Instead, you should just check cookie _as a fallback_ in the place where we check the header: https://github.com/makerdao-ses/switchboard-boilerplate/blob/40c536bd2fa1acb621baa722c3fbd6c53c3ca65f/api/src/graphql/context.ts#L29

valiafetisov: The other reason why it's not a good idea: globalHeaders will be set only once, when the graphql playground is requested from the server. If the user signs out / signs in as a different user it will still stay the same

KirillDogadin-std: > which does not fulfil the requirement defined in the issue

to be fair, this was listed as a suggestion of implementation

> Instead, you should just check cookie as a fallback in the place where we check the header:

well, this effectively does exactly this:
  - if the header is provided, it uses it
  - if the header is not provided, it uses the cookie one.

> If the user signs out / signs in as a different user it will still stay the same

if done in the single browser tab - this is not true. Since the button for the sign out is located in the other tab, on return to the playground the cookie will be reset already with the new user.

if done in different browser tabs or via playground: there might be problems indeed (if the second tab with playground is not refreshed), but wouldn't this also be a problem for the in-place check? 

Moreover, since the playground does not send the cookies automatically, i would have to tweak how the request goes through the iframe to send the cookie header along with it. Is it really worth it? Asking since i don't see what is another user story (single tab) would result into an edgecase where my suggested solution does not work.

valiafetisov: > to be fair, this was listed as a suggestion of implementation

Sorry, then I shouldn't call this a requirement, but I expect that if implementation suggestion is not followed, there is a written argument why the other solution is better 🙂 

> if the header is not provided, it uses the cookie one

Interesting, was it confirmed via functional testing? Because I can't guess it's the current behaviour by looking at the code, since you always set header even if header is not present.

>  Since the button for the sign out is located in the other tab, on return to the playground the cookie will be reset already with the new user because the playground is re-requested.

Not sure what do you mean by the "tab" in this context, but I'm talking about the scenario where the login is happening in a "different _browser_ tab" than the playground is open

> but wouldn't this also be a problem for the in-place check?

I imagine the browser should send updated cookies with every new request, since cookies are shared between the tabs

> since the playground does not send the cookies automatically

That's interesting, what do you have modify to send them? And what iframe have to do with it?

valiafetisov: > Moreover, since the playground does not send the cookies automatically
> well, i did not find a way to configure the playground in-place

Wouldn't this work? https://github.com/graphql/graphql-playground/issues/1073#issuecomment-526509499

> In this case one has to just hit refresh page.

But the whole point of this PR is to improve UX of the logged in user (:

> To address the edgecase i would simply suggest keeping track of the me endpoint and reloading the playground when me stops returning 200.

If the solution above doesn't work. Instead of your proposal, I would suggest to add a watcher to the `graphql-playground` vue code and then refresh contents if it changes of the `iframe` element, [eg](https://stackoverflow.com/a/86771)

KirillDogadin-std:
tested manually in diff tabs - seems to work out now.

valiafetisov (approved):

PR #53, opened 2023-05-02
Wallet auth

### LINKED ISSUE #52: Implement login via eth wallet
### Goal
User is able to login via eth wallet

### Context
Instead of using login + password implemented initially (based on the initial requirements), we should aim to support blockchain-based authentication as the primary one. Therefore, login + password authentication have to be removed and instead a new authentication methods have to be introduced. On the technical level: we would ask the user to sign an off-chain message (e.g. sessionId) using their wallet (e.g. metamask), then confirm that signed message actually signed by the address that they claim to own. The result of the procedure will be (as far as I understand the procedure):
- Address of the user in the `eth:0x...` format (instead of the login) stored in the database
- Message signed by this address, stored in the database (to be able to audit this transaction in the future)
- The rest can stay the same for now (sessions, jwt tokens, etc) as it has to be discussed first

My current, uneducated understanding:
- 1st endpoint creates "challenge" (ie a message to be signed) stores it in the database
- 2nd endpoint accepts the address and the challenge result
  - verifies that message is signed by the claimed address
  - returns valid token

### Tasks
- [ ] Find the official spec that describes the authentication process using ethereum public/private keys
  - [ ] Link it here
  - [ ] Make short implementation summary if needed
- [ ] Find and compare established libraries that implement the spec
- [ ] Outline implementation proposal
- [ ] Implement API methods
  - [ ] Provide clear instructions on how to use it without UI inside a readme
- [ ] Update UI to support it

KirillDogadin-std:
will close #52 

posted on request.


api/prisma/schema.prisma
```diff
@@ -21,8 +21,6 @@ model CoreUnit {
 
 model User {
   id       String @id @default(uuid())
```
valiafetisov, line 23: The primary identity of the user should be the eth address in the format `eth:0x...`

api/prisma/schema.prisma
```diff
@@ -41,3 +39,7 @@ model Session {
 
   @@unique([createdBy, id])
 }
+
+model PendingAuth {
+  id String @id @default(uuid())
+}
```
valiafetisov, line 45: I suggest to add creation date (to be able to expire or delete challenges)

api/prisma/schema.prisma
```diff
@@ -41,3 +39,7 @@ model Session {
 
   @@unique([createdBy, id])
 }
+
+model PendingAuth {
```
valiafetisov, line 43: Let's call it `Challenge` or `SessionChallenge` to improve consistency

api/src/modules/Session/model.ts
```diff
@@ -74,6 +82,51 @@ const generateTokenAndSession = async (
   };
 };
 
+async function createUserIfNotExists(prisma: PrismaClient, wallet_address: string) {
```
valiafetisov, line 85: Shouldn't this be part of the User model?

api/src/modules/Session/model.ts
```diff
@@ -74,6 +82,51 @@ const generateTokenAndSession = async (
...
+    return user
+  }
+  return await prisma.user.create({
+    data: {
+      id: wallet_address,
+    }
+  })
+
+}
+async function createPendingSession(prisma: PrismaClient): Promise<string> {
```
valiafetisov, line 101: - Please improve consistency by calling it the same way
- Please improve consistency by returning the same type as the name suggests
```suggestion
async function createChallenge(prisma: PrismaClient): Promise<Challenge> {
```

api/src/modules/Session/model.ts
```diff
@@ -74,6 +82,51 @@ const generateTokenAndSession = async (
...
+async function verifyMessageAndCreateSession(
+  prisma: PrismaClient,
+  message: string,
+  signature: string,
+  session: { expiryDurationSeconds?: number | null; name: string },
+  isUserCreated: boolean = false
+) {
+  const siweMessage = new SiweMessage(message);
+  const userId = siweMessage.address;
+  const sessionId = siweMessage.nonce;
```
valiafetisov, line 120: The whole point of `nonce` is to validate that it exists in our database, but currently you're not doing that. As you don't delete used challenges from our database the replay attack would still be possible, even if you would validate it. Moreover you're misleading by calling it `sessionId`, since it's not yet a session, it's a challengeId.

- [ ] Make sure that this function throws error if nonce is unknown to us
- [ ] Make sure to delete the challenge once it used once
- [ ] Make sure to do everything in one sql transaction

api/src/modules/User/resolvers.ts
```diff
@@ -12,24 +12,21 @@ export const me = queryField('me', {
...
-  },
-  resolve: async (_parent, { user: userNamePass }, ctx) => {
-    const { id } = await ctx.prisma.user.getUserByUsernamePassword(userNamePass);
-    return ctx.prisma.session.createSignInSession(id);
-  },
+export const getChallenge = mutationField('getChallenge', {
+  type: 'Challenge',
+  resolve: async (_, __, ctx) => {
+    return ctx.prisma.session.getChallenge();
+  }
```
valiafetisov, line 19: Instead of returning a single id from this endpoint and ask users to generate complete message on their own, I would return a complete message to be singed as a basic string

valiafetisov (commented):
Once points are addressed, can you please provide instructions on how to test it using playground/cli?

valiafetisov:
Closing in favour of https://github.com/powerhouse-inc/switchboard-boilerplate/pull/55 (that uses a slightly different approach)

PR #55, opened 2023-06-20
Wallet auth

### LINKED ISSUE #52: Implement login via eth wallet

valiafetisov:
Closes #52


valiafetisov:
Current state: API service have fully working sign in with wallet functionality (manually testable). Next steps:
- [ ] Update tests to cover new API logic
- [ ] Update frontend to use "Sign in with Ethereum" button

### Testing scenario
1. Create challenge via executing `createChallenge` graphql mutation
    - Example request (that have to contain your public etherium address)
        ```gql
        mutation {
            createChallenge(
                address: "paste_your_ethereum_address"
            ) {
                nonce
                message
            }
        }
        ```

    - Example response (that contains hex-encoded message)
        ```json
        {
            "data": {
                "createChallenge": {
                    "nonce": "6f4c7f7cd61a499290e68a2740957407",
                    "message": "0x6c6f63616c686f73742077616e747320796f7520746f207369676e20696e207769746820796f757220457468657265756d206163636f756e743a0a3078333139386235354434393732393346373333463236333137353563373636653131636535363866350a0a0a5552493a20687474703a2f2f6c6f63616c686f73743a333030300a56657273696f6e3a20310a436861696e2049443a20310a4e6f6e63653a2036663463376637636436316134393932393065363861323734303935373430370a4973737565642041743a20323032332d30362d32305431373a34353a35362e3435385a"
                }
            }
        }
        ```

2. Sign provided message
    - Either [using your metamask wallet](https://docs.metamask.io/wallet/how-to/use-siwe/)
        ```js
        // this should be executed at the browser console of the graphql playground
        await ethereum.request({
            method: 'personal_sign',
            params: [
                'paste_message_from_above',
                'paste_your_ethereum_address'
            ]
        });
        ```

    - Or using foundry command line tool called `cast` (note: you will be asked for your private key; for other auth methods, [read the cli docs](https://book.getfoundry.sh/reference/cast/cast-wallet-sign))
        ```sh
        $ cast wallet sign -i "message_from_above"
        ```

3. Provide signature back to the API to get usual JWT token back
    - Example request
        ```gql
        mutation {
            solveChallenge(
                nonce: "paste_nonce_from_above"
                signature: "paste_signature_generated_in_step_2"
            ) {
                token
            }
        }
        ```

    - Example response
        ```json
        {
            "data": {
                "solveChallenge": {
                    "token": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzZXNzaW9uSWQiOiI5ZGM1NjI3Mi1hMjBjLTRmM2YtYjM5MC1kZDc2NjE1NTA0YTYiLCJpYXQiOjE2ODczMzc2MDEsImV4cCI6MTY4Nzk0MjQwMX0.z1lJlKXnCbcex59JkU9j7hfRGhR2EBrnUE8phwPN7C0"
                }
            }
        }
        ```

4. Use provided JWT token to make subsequent API requests
    - Either sent as `Authorization: Bearer paste_your_token_from_above`
    - Or set as `Authorization` cookie

LukSteib (approved):
Approved based on screenshared review. What we might still adjust: 
- [ ] Make it work in the docker based setup

api/README.md
```diff
@@ -31,10 +31,11 @@ Some environment variables are pre-configured for the development. You can copy
 
 - `DATABASE_URL` (required): path to the database file
 - `JWT_SECRET` (required): server's jwt secret
-- `PORT` (optional, default: 3000): port on which the server will run
-- `AUTH_SIGNUP_ENABLED` (optional, default: `false`): if signing up mutation is allowed (i.e. user creation via endpoint is enabled)
-- `JWT_EXPIRATION_PERIOD` (optional, default: `'7d'`): how soon the signed jwt token will expire
-- `DEBUG` (optional): if set, enables the different more explicit logging mode where debug levels are set to `debug` for the app's logger and `query` for db logger
+- `PORT` (optional, default `3000`): port on which the server will run
+- `API_ORIGIN` (optional, default `http://0.0.0.0:${PORT}`): the URL at which the API is running. it's important to provide this variable in production since it influences the message signed during authorization
+- `AUTH_SIGNUP_ENABLED` (optional, default: `false`): if signing up is allowed. In case it's not set, no new users can be created, but old users can still sign up
```
KirillDogadin-std, line 36: nit
```suggestion
- `AUTH_SIGNUP_ENABLED` (optional, default: `false`): if signing up is allowed. In case it's not set, new users _cannot_ be created, but old users _can_ still sign up
```

api/src/modules/Session/model.ts
```diff
@@ -1,57 +1,48 @@
...
-    t.date('referenceExpiryDate');
-    t.nonNull.string('referenceTokenId');
-    t.nonNull.boolean('isUserCreated');
-    t.string('name');
-    t.date('revokedAt');
-    t.string('allowedOrigins');
-  },
-});
+export function getSessionCrud(
+  prisma: Omit<PrismaClient, '$connect' | '$disconnect' | '$on' | '$transaction' | '$use'>,
```
KirillDogadin-std, line 8: If this one type is used multiple times (>1) - should it be extracted and defined separately?

api/src/modules/User/model.ts
```diff
@@ -1,76 +1,30 @@
...
-      } catch (e: any) {
-        if ('code' in e && e.code === 'P2002') {
-          throw new GraphQLError('Username already taken', { extensions: { code: 'USERNAME_TAKEN' } });
-        }
-        /* istanbul ignore next @preserve */
-        throw e;
-      }
-      return createdUser;
+      logger.error('Upserted user', usertedUser);
+      return usertedUser;
```
KirillDogadin-std, line 26:
```suggestion
      return upsertedUser;
```

frontend/composables/useAuth.ts
```diff
@@ -35,31 +32,44 @@ const useAuth = function () {
     }
   }
 
-  const signIn = async (username: string, password: string) => {
-    const { data, error } = await useAsyncGql('signIn', { username, password })
-    if (error.value || !data.value?.signIn?.token) {
+  const createChallance = async function (address: string) {
```
KirillDogadin-std, line 35: typo?

api/src/modules/Session/model.ts
valiafetisov: Found suitable type inside prisma namespace

SidestreamSweatyPumpkin (approved):
LGTM

LukSteib (approved):
- Re-checked via docker and worked as expected ✔️ 
- Checked reference implementation ([here](https://login.xyz/)) - no further feedback based on that, so all good ✔️ 

One remaining question (non approval blocking):
Tried browser session without metamask extension and received the following error:
```
Sign in error: could not coalesce error (error={ "message": "Cannot read properties of undefined (reading 'request')" }, code=UNKNOWN_ERROR, version=6.6.1)
```
Assume this is expected, right?

valiafetisov:
> Assume this is expected, right?

No, I actually haven't tested it. [Updated the code](https://github.com/powerhouse-inc/switchboard-boilerplate/pull/55/commits/5d07ba08023a0b7cdabc6b8ac8cb70cba4027868#diff-b8c4f5a6adaf24caa107129924728d86a721bbc1bd6e081b3bb2571b77ce6a04) with a better error message "Please install web-based wallet such as Metamask".

KirillDogadin-std (approved):

LukSteib (changes requested):
- [ ] Still getting the following error message: `Sign in error: could not coalesce error (error={ "message": "Cannot read properties of undefined (reading 'request')" }, code=UNKNOWN_ERROR, version=6.6.1)` -> not the one from latest commit

LukSteib (approved):
Just didn't recreate the docker images in the first place :( 
My bad... 

Updated error message works ✔️ 

Approved

PR #58, opened 2023-06-29
Improve documentation integration

### LINKED ISSUE #57: Improve documentation integration
### Goal
Implement dynamic `document-model-libs` documentation

### Context
As all changes [proposed in the initial issue](https://github.com/powerhouse-inc/switchboard-boilerplate/issues/22#issuecomment-1508379896) were [reported as addressed](https://github.com/powerhouse-inc/switchboard-boilerplate/issues/22#issuecomment-1560803749), it's now our task to update the integration and make it dynamic (so whenever `document-model-libs` is updated, documentation is also updated automatically)

### Tasks
- [x] Add `document-model-libs` as npm package
- [x] Update [`generateDocs` script](https://github.com/powerhouse-inc/switchboard-boilerplate/blob/main/frontend/scripts/generateDocs.ts)
- [ ] Remove initial static md files
- [x] Manually test generation via docker-compose (as it might require additional changes to take md files from `node_modules`)

KirillDogadin-std:
close #57


KirillDogadin-std:
@valiafetisov 

not sure what the requirement `remove initial static md files` means in the issue. Could you elaborate what is wrong?

KirillDogadin-std:
Docs are gitignored now and are generated automatically after install , after update, before `dev`, before `start`. That should cover the most cases of the tool used.

valiafetisov (commented):
- [ ] Is it possible to display `DOCS_VERSION` somewhere at the `/documentation` page as well?

frontend/scripts/generateDocs.ts
```diff
@@ -1,24 +1,41 @@
...
   fs.writeFileSync(outputFilePath, content, { encoding: 'utf8' })
 }
 
 function main () {
-  if (!DOCS_DIR) {
-    throw new Error('Please provide DOCS_DIR env variable which points to documentation directory')
+  const dirname = path.resolve(__dirname, '..', 'content', 'documentation')
+  const outputFilePath = path.resolve(dirname, 'index.md')
+  const outputVersionPath = path.resolve(dirname, '.version')
+  if (!fs.existsSync(outputVersionPath) || fs.readFileSync(outputVersionPath, { encoding: 'utf8' }).toString() !== DOCS_VERSION) {
```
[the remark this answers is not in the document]
KirillDogadin-std: We only want to regenerate documentation if it does not exist or if the package version is not the same as the one the documentation is generated for. 

It does not seem reasonable to always regenerate the docs.

frontend/scripts/generateDocs.ts
```diff
@@ -1,24 +1,41 @@
...
+const PACAKGE_NAME = '@acaldas/document-model-libs'
+
+try {
+  execSync(`npm ls ${PACAKGE_NAME}`)
+} catch (e) {
+  throw new Error(`Failed to generate documentation from the dependency (${PACAKGE_NAME}), did you run \`npm install\`?`)
+}
+
+// eslint-disable-next-line import/first
 import { concatMdSync } from 'concat-md'
```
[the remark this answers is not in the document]
KirillDogadin-std: > Why was the import moved down here?

Since the generation runs before `npm run dev` - it is required to cover for the case where the `npm install` was not executed. I.e. if one removes `node_modules/` and runs a command that triggers generation - it has to be clear what is wrong.

> You know if wouldn't influence the order of events and concatMdSync will be imported before execSync is executed anyway, right?

I assume there's an error in the question and it is actually `You know it...` instead of `if...`. 

I disagree on the basis of the following evidence. 

This is the error if all the imports grouped together (`concat-md` is in the same batch as the others).

```

> powerhouse-switchboard-frontend@0.0.1 generateDocs
> vite-node ./scripts/generateDocs.ts

Failed to load url concat-md (resolved id: concat-md) in /Users/kirilldogadin/work/switchboard-boilerplate/frontend/scripts/generateDocs.ts. Does the file exist?
file:///Users/kirilldogadin/node_modules/vite/dist/node/chunks/dep-ca21228b.js:39470
        const err = new Error(`Failed to load url ${url} (resolved id: ${id})${importer ? ` in ${importer}` : ''}. ${msg}`);
                    ^

Error: Failed to load url concat-md (resolved id: concat-md) in /Users/kirilldogadin/work/switchboard-boilerplate/frontend/scripts/generateDocs.ts. Does the file exist?
    at loadAndTransform (file:///Users/kirilldogadin/node_modules/vite/dist/node/chunks/dep-ca21228b.js:39470:21) {
  code: 'ERR_LOAD_URL'
}
```

and this is the error if i run the generation as it is now

```
> powerhouse-switchboard-frontend@0.0.1 generateDocs
> vite-node ./scripts/generateDocs.ts

/Users/kirilldogadin/work/switchboard-boilerplate/frontend/scripts/generateDocs.ts:11
  throw new Error(`Failed to generate documentation from the dependency (${PACAKGE_NAME}), did you run \`npm install\`?`);
        ^

Error: Failed to generate documentation from the dependency (@acaldas/document-model-libs), did you run `npm install`?
    at /Users/kirilldogadin/work/switchboard-boilerplate/frontend/scripts/generateDocs.ts:10:9
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at ViteNodeRunner.directRequest (file:///Users/kirilldogadin/node_modules/vite-node/dist/client.mjs:308:5)
    at ViteNodeRunner.cachedRequest (file:///Users/kirilldogadin/node_modules/vite-node/dist/client.mjs:152:14)
    at ViteNodeRunner.executeFile (file:///Users/kirilldogadin/node_modules/vite-node/dist/client.mjs:132:12)
    at CAC.run (file:///Users/kirilldogadin/node_modules/vite-node/dist/cli.mjs:70:5)
```

P.S.

You also forgot to add a prooflink to this statement :D

frontend/content/documentation/index.md
valiafetisov, line 1: Despite you adding this file to gitignore, it's still present in git until you `git rm` it. Or is it desirable?

frontend/components/layout/TheMenu.vue
```diff
@@ -15,11 +15,14 @@ import {
...
 
+const config = useRuntimeConfig()
+const version = config.public.DOCUMENTATION_VERSION
+
 const links = [
   {
     id: '/documentation',
     icon: DocumentText24Regular,
-    label: 'Documentation'
+    label: `Documentation v${version}`
```
valiafetisov, line 25: Not sure it's a good idea to add it to the header, but I would let @LukSteib decide how it looks and if it better to move to the page itself

frontend/scripts/generateDocs.ts
```diff
@@ -1,24 +1,41 @@
...
+  execSync(`npm ls ${PACKAGE_NAME}`)
+} catch (e) {
+  throw new Error(`Failed to generate documentation from the dependency (${PACKAGE_NAME}), did you run \`npm install\`?`)
+}
+
+// eslint-disable-next-line import/first
 import { concatMdSync } from 'concat-md'
 
-const DOCS_DIR = process.env.DOCS_DIR
+const DOCS_DIR = path.resolve(__dirname, '..', 'node_modules', ...PACKAGE_NAME.split('/'), 'dist', 'docs')
```
valiafetisov, line 16: Nit: you actually don't need to split paths into chunks, you only need to 1) avoid joining strings via `+` and 2) wrap them into some `path` method, which will take care of setting right delimiter per platform, [see example in the docs](https://nodejs.org/api/path.html#pathjoinpaths)
```suggestion
const DOCS_DIR = path.resolve(__dirname, '../node_modules', PACKAGE_NAME, 'dist/docs')
```

frontend/scripts/generateDocs.ts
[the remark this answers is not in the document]
valiafetisov: Then, I would also move writing of `DOCS_VERSION` into the `generateMdDocs` function. Otherwise, now, the version file can be updated without updating the docs

frontend/content/documentation/index.md
KirillDogadin-std: fixed

frontend/components/layout/TheMenu.vue
KirillDogadin-std: waiting for @LukSteib

KirillDogadin-std: if not in the header - then pls tell the spot so i don't juggle this version tag 3 times before it lands correctly.

LukSteib: Also wouldn't want to add it to the header. 
Instead let's use the page itself and just add it on top of the first heading (ie. `Readme` in screenshot below) in a decent way. Doesn't have to be super prominent but good to include

image

valiafetisov: I would suggest to keep it on the grey background right on top of the "README", to use the same font size and color as sidebar text and to say `Documentation of the package version ${version}`

KirillDogadin-std: image

valiafetisov: > "on the grey background"

The above [suggestion](https://github.com/powerhouse-inc/switchboard-boilerplate/pull/58#discussion_r1246829380) was to not put the text on the white background to indicate that it is not part of the readme itself. If you did not follow the suggestion on purpose, please provide a reason, otherwise I assume you missed it 😞

KirillDogadin-std: Ah, sorry, i misunderstood the requirement. My brain for some reason imagined that the letters are "transparent" and there's a grey background behind them 🤦🏻

KirillDogadin-std: image

valiafetisov: Desired state

desired-state

KirillDogadin-std: image

frontend/package.json
```diff
@@ -36,6 +39,7 @@
     "vue-tsc": "^1.0.24"
   },
   "dependencies": {
+    "@acaldas/document-model-libs": "^1.0.15",
```
valiafetisov, line 42: One thing that I haven't noticed before: this package actually have to be added to the `api` service instead of the frontend (since it will be used there in the first place) – otherwise we again may end up in the same situation where package version which is used by the api is different from the documentation displayed on the frontend

frontend/package.json
```diff
@@ -10,12 +10,15 @@
...
+    "predev": "npm run generateDocs",
     "dev": "nuxt dev",
     "generate": "nuxt generate",
     "preview": "nuxt preview",
-    "postinstall": "nuxt prepare",
+    "postinstall": "nuxt prepare && npm run generateDocs",
+    "postupdate": "npm run generateDocs",
     "lint": "eslint --max-warnings 0 .",
     "typecheck": "nuxt typecheck",
+    "prestart": "npm run generateDocs",
```
valiafetisov, line 21: Based on the situation described above and to simplify dev experience (not requiring them to `cd api && npm i && cd ../frontend`), I would suggest to:
- `A` only generate docs inside docker (so probably only keep `prestart` here and remove hard requirement for the content/md file to run `npm run dev`)
- `B` use something like https://www.npmjs.com/package/live-plugin-manager to install / download package from the `api/package.json` locally – this is more like a crazy hack, so I wouldn't recommend it

KirillDogadin-std: 1. Frontend docker file already copies the api folder 
  - i've added the `file:` dependency in the `package.json` of frontend.
2. The Frontend dockerfile does not use `npm start` instead it goes with `
CMD [ "node", "/app/frontend/.output/server/index.mjs" ]
`
    - therefore i went with `postbuild`

frontend/package.json
KirillDogadin-std: It does not make 100% sense to me.

If the dependency is in the api , then the api has to generate and serve the readme file - correct?

This then implies that i would have to write logic that handles this: 
  - provides the readme in some format via the api 
  - gets the readme from the api and displays this on the frontend.

Correct? Or did you have something else in mind?

KirillDogadin-std: Ok, now the dep is in the api. Docker-compose works for me.

frontend/package.json
KirillDogadin-std: Dockerfile for frontend does not use `npm start` so i've added the documentation generation on `prebuild`

KirillDogadin-std:
pls try :

1. removing `node_modules/` in both `frontend` and `core` directories.
2. running `npm i` in both of the directories.
3. docker-compose again.

LukSteib (changes requested):
- [ ] After guidlines above docker starts now. However, except for version indicator no documentation is rendered anymore. Assume this is not expected... 

image

KirillDogadin-std:
aaalright. I am failing to reproduce:

1. freshly clone repo
2. run `docker-compose up --build --force-recreate`

---

And the documentation is displayed. 

Please:

1. ensure it does not work on the freshly cloned repo
2. if it does not - post complete list of steps you've taken to get to the broken state (terminal commands and their output)

LukSteib (approved):
1. clone fresh repo
2. run `npm i` in `/api` and `/frontend` folders
3. generate `.env` file via `cp ./api/developer.env ./api/.env`
4. run `docker-compose up --build --force-recreate

-> Worked ... LGTM

frontend/README.md
```diff
@@ -25,8 +25,15 @@ Endpoint available at `/healthz` path. Provides response if frontend is currentl
...
 
 One of the frontend's features is displaying the documentation about document model.
-The documentation is provided to the service externally as raw typedoc output in markdown format.
+The documentation is provided to the service externally as part of the npm package.
 
 To display the documentation on the frontend the provided data has to be processed and represented in the form of a single file.
 
-TODO: add the precise process in separate issue
+The documentation gets updated or generated along with the package initially containing it: on install or update (postinstall and postupdate hooks).
+The documentation gets updated or generated before `npm run dev` and `npm run start`.
```
valiafetisov, line 33: This is no longer true, right?

frontend/pages/documentation.vue
```diff
@@ -57,8 +61,13 @@ useHead({
...
       </div>
-      <div class="flex-grow !w-[calc(100%-18rem)] p-4 pt-2 mt-4 bg-white w-full">
-        <ContentRendererMarkdown v-if="documentation" :value="documentation" class="markdown-body w-full" />
+      <div class="flex-grow !w-[calc(100%-18rem)] mt-4 w-full">
+        <div v-if="documentationLabel" class="pb-2">
+          <span class="text-gray-400"> {{ documentationLabel }} </span>
+        </div>
+        <div class="bg-white p-4">
+          <ContentRendererMarkdown v-if="documentation" :value="documentation" class="markdown-body w-full" />
+        </div>
```
valiafetisov, line 70: One quick suggestion to fix the page if the frontend is being run outside of the docker is to remove whitespace (which otherwise makes page look broken)

Screenshot 2023-07-05 at 17 08 21

```suggestion
        <div v-if="documentation" class="bg-white p-4">
          <ContentRendererMarkdown :value="documentation" class="markdown-body w-full" />
        </div>
```

The other suggestion, is to show something like `Follow the instructions in the frontend/README to generate the docs` (but probably only in development mode?)

valiafetisov (commented):
> 2. run `npm i` in `/api` and `/frontend` folders

This shouldn't be a pre-requirement, since there is no preinstallation of npm packages available in CI/CD. But I've tested it on the freshly cloned repo without step `2.` and it also properly generates documentation page

valiafetisov (approved):

PR #67, opened 2023-09-20
Make API extensible via npm packages

KirillDogadin-std:
Closes https://github.com/powerhouse-inc/switchboard-boilerplate/issues/61

Screenshot 2023-11-08 at 11 21 54


KirillDogadin-std:
@valiafetisov time for opinion

If the module is conform with nexus structure, then it's quite easy to just define it and import to the schema. In more sofishticated cases when we e.g. need to define crud operations in `models.ts`. See the tricks that i've used to make it work together.

module-example/src/model.ts
valiafetisov, line 1: Let's remove this package (which reimplements Challenge) because it overly complex for an example

module-example-2/src/model.ts
```diff
@@ -0,0 +1,16 @@
+import { objectType } from 'nexus/dist';
+
+export const CoreUnit = objectType({
+  name: 'CoreUnit',
```
valiafetisov, line 4: Why redefining the same type here doesn't break typescript or nexus?

api/src/modules/index.ts
```diff
@@ -1,4 +1,2 @@
 export * from './User';
-export * from './CoreUnit';
```
valiafetisov, line 2: Instead of moving a specific module into the package, please define a new `ExampleType` with actual example resolver (which for example can return a number of users in the database)

module-example/package.json
```diff
@@ -0,0 +1,59 @@
...
+        "nexus-validation-plugin": "^0.2.0",
+        "pino": "^8.11.0",
+        "pino-http": "^8.3.3",
+        "pino-pretty": "^10.0.0",
+        "siwe": "^2.1.4",
+        "vite-node": "^0.29.2",
+        "wildcard-match": "^5.1.2",
+        "zod": "^3.21.4"
+    },
+    "devDependencies": {
```
valiafetisov, line 41: Do you actually need all of those dependencies and devDependencies? Otherwise, please clean them up

api/src/setupPlugins.ts
```diff
@@ -0,0 +1,14 @@
+import { setup } from 'powerhouse-switchboard-plugin-demo';
```
valiafetisov, line 1: Why does one plugin requires calling a setup when the other does not?

api/src/graphql/schema.ts
```diff
@@ -4,10 +4,12 @@ import { validationPlugin } from 'nexus-validation-plugin';
 import { applyMiddleware } from 'graphql-middleware';
 import * as types from '../modules';
 import { GQLDateBase } from './dateSchema';
+import * as demoTypes from 'powerhouse-switchboard-plugin-demo';
+import * as anotherDemoTypes from 'powerhouse-switchboard-plugin-demo-2';
```
valiafetisov, line 8: This looks like magic and doesn't give me (or for sure any other developer) a sense of what needs to be exported from the package to make it work? Can you please avoid using `*`? 

Ideally, as I proposed in the issue `setup()` (which is currently called in the other file) would return required types (so you import something from the package once and call it once as well)

api/src/setupPlugins.ts
```diff
@@ -0,0 +1,14 @@
...
+import { getSessionCrud } from './modules/Session'
+import {API_ORIGIN} from './env'
+
+export const setupAll = () => {
+  setup({
+    apiOrigin: API_ORIGIN,
+    crud: {
+      getUserCrud,
+      getSessionCrud
+    }
```
valiafetisov, line 12: Can't you have this crud object already "prebuilt" if you will provide `prisma` to the setup?

api/src/database.ts
```diff
@@ -1,8 +1,11 @@
 import { PrismaClient, Prisma } from '@prisma/client';
 import { Level as PinoLevel } from 'pino';
 import { getChildLogger } from './logger';
-import { getUserCrud, getSessionCrud, getChallengeCrud } from './modules';
+import { getUserCrud, getSessionCrud } from './modules';
+import { getChallengeCrud } from 'powerhouse-switchboard-plugin-demo'
```
valiafetisov, line 5: This is exactly what I want to prevent: that packages export multiple different functions that are then have to be manually put together in the main app

But removing `Challenge` module should already decrease complexity for this PR. As importing module that will provide functionality to other modules is too much

module-example-2/src/model.ts
KirillDogadin-std: it is not exported from the `api/modules` (look up index.ts) changes. So there is no conflict since only one place (the module-example directory) has this exported.

api/src/setupPlugins.ts
KirillDogadin-std: - one plugin has to have some crud defined already, another one does not.
- one plugin relies on the api_origin env var, another one does not.

valiafetisov: So the suggestion is to make this interface universal: all plugins have to return `setup` function, which will receive the same context object with env vars etc. If a plugin doesn't need the context, it just doesn't use it.

module-example-2/src/model.ts
valiafetisov: But would it throw an error if it exports the same type?

api/src/setupPlugins.ts
KirillDogadin-std: it should be possible, although requires some alterations in the original concept where `getCrud` functions are provided to the `getChallengeCrud` function. Since the challenge related functionality will be removed from the pr - this discussion point seems less relevant as of now. But generally i would just orient the direction of the decision so that the types would be easier to handle.

KirillDogadin-std:
did the independent counter example, the gql query for it:

```
query {
  countUsers(message: "aslekjf") { count, message}
}
```

api/src/database.ts
KirillDogadin-std: see the updated version

api/src/graphql/schema.ts
KirillDogadin-std: see the change.

api/src/setupPlugins.ts
KirillDogadin-std: see the current version

module-example/package.json
KirillDogadin-std: cleaned up according to `depcheck`

api/src/modules/index.ts
KirillDogadin-std: see the adjustment

module-example-2/src/model.ts
KirillDogadin-std: if on [ae7d149](https://github.com/powerhouse-inc/switchboard-boilerplate/pull/67/commits/ae7d14910407fba42b1dc977778a459e07c02996) i change the name of the type from `Counter` to `CoreUnit` - no error is thrown and it seems like it just overwrites the logic there.

Although due to redefinition above the typecheck command throws an error for a resolver for the actual core unit. So seems like the typesystem in the CI is still able to catch the conflicting behaviour.

module-example/src/model.ts
KirillDogadin-std: done

api/package.json
```diff
@@ -35,6 +35,7 @@
         "pino": "^8.11.0",
         "pino-http": "^8.3.3",
         "pino-pretty": "^10.0.0",
+        "powerhouse-switchboard-plugin-demo": "file:../module-example",
```
valiafetisov, line 38: Can you please make it consistent: either call both things `plugin-demo` or `module-example`. I suggest to use `module-example` 🙂

api/src/generated/nexus.ts
```diff
@@ -75,6 +75,10 @@ export interface NexusGenObjects {
     name?: string | null; // String
     shortCode?: string | null; // String
   }
+  Counter: { // root type
+    count: string; // String!
+    message: string; // String!
+  }
```
valiafetisov, line 81: Maybe we would then need to remove those generated files from git whatsoever, to avoid update to our code when a module is updated. Is it even possible? (to remove `generated` folder from repo)

module-example/package.json
```diff
@@ -0,0 +1,28 @@
...
+        "@prisma/client": "^4.11.0",
+        "ethers": "^5.7.2",
+        "graphql": "^16.7.1",
+        "nexus": "^1.3.0",
+        "siwe": "^2.1.4"
+    },
+    "devDependencies": {
+        "@types/node": "^18.14.2",
+        "eslint": "^8.35.0",
+        "prisma": "^4.11.0",
```
valiafetisov, line 25: Why do you need `prisma`, `@prisma/client`, `graphql`, `ethers` and `siwe` in the package? Do you use/import anything from there?

api/src/graphql/schema.ts
```diff
@@ -4,10 +4,13 @@ import { validationPlugin } from 'nexus-validation-plugin';
 import { applyMiddleware } from 'graphql-middleware';
 import * as types from '../modules';
 import { GQLDateBase } from './dateSchema';
+import {setup} from 'powerhouse-switchboard-plugin-demo';
+
+const demoTypes = setup();
```
valiafetisov, line 9: As suggested previously:
- [ ] Please provide a single separate config file where only modules are imported and setup functions are added to some array, which is the only action that needs to be done by the developer
- [ ] Please provide example of adding more data to the plugin (such as origin you've used previously, crud actions)

module-example/package.json
KirillDogadin-std: cleaned up

api/src/graphql/schema.ts
KirillDogadin-std: Ok, got the proposable structure. See if it's ok. Apart from this typechecks just explode with errors because there's no typing as of now.

KirillDogadin-std:
Currently trying to find a way of writing the module so that the type system remains intact.
To discuss this issue i would assume that code demo should be available, hence will open a separate draft PR that is not for merging but for demo purposes only.

api/src/graphql/schema.ts
```diff
@@ -28,4 +31,6 @@ export const schema = makeSchema({
     export: 'Context',
   },
 });
+
+// @ts-ignore
```
valiafetisov, line 35: Why is it here? Please prefer to [specify exact rules](https://stackoverflow.com/a/43620820/2270613) that needs to be ignored

module-example/prisma/extesion.ts
```diff
@@ -0,0 +1,17 @@
+import { Prisma } from '@prisma/client';
+
+export const prismaExtension = Prisma.defineExtension((client: any) => client.$extends({
```
valiafetisov, line 3: I think prisma client is this magical object that is pre-built by the `prisma generate` command based on the prisma schema. So to avoid `any` and use correct version of the client, you need to import a different prisma client. But maybe this introduces unnecessary interdependency, which we can sacrifice over type safety. I have to try/think about it

valiafetisov: I would leave it for now to not introduce interdependency

valiafetisov (approved):

valiafetisov:
I've added a readme section on how to make new modules: https://github.com/powerhouse-inc/switchboard-boilerplate/blob/b863e6a2b778fe572298123caa981cc1d33b8ecf/api/README.md#adding-new-modules

PR #68, opened 2023-10-05
[DO NOT MERGE] Modularization prisma type compatability playground

KirillDogadin-std:
related to #67

### DIFF
--- a/api/src/database.ts
+++ b/api/src/database.ts
@@ -1,7 +1,7 @@
 import { PrismaClient, Prisma } from '@prisma/client';
 import { Level as PinoLevel } from 'pino';
 import { getChildLogger } from './logger';
-import { getUserCrud, getSessionCrud, getChallengeCrud } from './modules';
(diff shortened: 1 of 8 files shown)


KirillDogadin-std:
__(1)__

At the current state the problem is that the generated extended types are:
  - `GetClient<PrismaClient<...>>` on the api side
  - `{} & {coreUnit: DynamicModelExtensionThis<Prisma.TypeMap<Args & {result: {}; model: {user {...` on the plugin side.

And both are generated via `prisma.$extend` functionality.

This seems suspicious and raises concern wether it will be ok to use multiple extends if we can't expect some specific type returned.

__(2)__

`GetClient` does not seem to be exported from the prisma package given that the prisma types are generated based on the same file.

---

I am currently looking for a way to get the api generated client to have just `PrismaClient` type and for alternative way of importing the `GetClient` type.

Given that the latest (as of now) POC state the service works without problems - this problem seems like ignorable and we could potentially just write modules in js instead of ts. Or in a leniant ts (with `: any`s)

PR #90, opened 2024-02-23
feat: next js frontend

froid1911:


ryanwolhuter (approved):
excellent refactor, great work.

I can see a few places for improvement but they should definitely be future PRs. This looks great for now.

We should probably refactor to use the new [/app directory](https://nextjs.org/docs/app), since that is becoming the best practice going forward. Server components will likely help simplify a lot of things here.

I have found life is a lot easier with [wagmi](https://wagmi.sh/). It brings a lot of convenience and offloads a lot of maintenance work.

wagmi now uses [viem](https://viem.sh/) under the hood. I have found it to be a _huge_ quality of life improvement over ethers.

PR #93, opened 2024-02-26
fix: deep links

froid1911:


frontend/src/app/graphql/[driveId]/page.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { useParams } from "next/navigation";
+import { Suspense } from "react";
+
+export default function GraphQLDrive() {
+  const params = useParams<{ driveId: string }>();
+
+  return (
+    <Suspense>
+      <GraphQLIframe
+        url={`${process.env.NEXT_PUBLIC_SWITCHBOARD_HOST}/explorer/${params?.driveId}`}
```
Comment (acaldas) [inline frontend/src/app/graphql/[driveId]/page.tsx:12]: Do we just want to use the current URL? We could use ´window.location´ to retrieve it

Comment (froid1911) [reply frontend/src/app/graphql/[driveId]/page.tsx:12]: would avoid that, because it makes it hard to develop when you only start frontend and database and not the entire stack (with backend and nginx proxy). /explorer points to backend via nginx.

the var is set on build process and default points to prod release on heroku

acaldas (approved):

PR #98, opened 2024-02-28
feat: internal listener

froid1911:


api/src/modules/document/listenerManager.ts
```diff
@@ -0,0 +1,71 @@
...
+
+function isListenerRegistered(drive: DocumentDriveDocument, listener: Listener) {
+  const listeners = drive.state.local.listeners;
+  return listeners.filter((l) => l.listenerId === listener.listenerId).length > 0;
+}
+
+async function registerListener(driveServer: DocumentDriveServer, driveId: string, listener: Listener) {
+  let drive = await driveServer.getDrive(driveId);
+
+  drive = reducer(drive, actions.addListener({ listener }));
```
acaldas, line 40: We need a better way to add operations. We also could have specific methods to add a listener ´addInternalListener(options, onTransmit)´

acaldas (approved):

PR #106, opened 2024-03-15
fix: await add fees

froid1911:


api/src/modules/real-world-assets/listener.ts
```diff
@@ -708,15 +708,16 @@ const surgicalOperations: Record<string, (input: any, portfolio: RWAPortfolio, p
 
     "ADD_FEES_TO_GROUP_TRANSACTION": async (input: AddFeesToGroupTransactionInput, portfolio: RWAPortfolio, prisma: Prisma.TransactionClient) => {
         logger.debug({ msg: "Adding fee transactions to group transaction", input });
+        // add fees
         for (const fee of input.fees ?? []) {
-            prisma.rWAGroupTransactionFee.create({
+            await prisma.rWAGroupTransactionFee.create({
```
acaldas, line 713: Should we do a createMany for all fees instead? Making it a single db operation

PR #170, opened 2024-05-02
fix: rwa portfolios query

froid1911:


api/src/modules/document/model.ts
```diff
@@ -50,21 +50,24 @@ export function getDocumentDriveCRUD(prisma: Prisma.TransactionClient) {
 
 
   let driveServer: DocumentDriveServer;
-  getRedisClient().then((redisClient) => {
```
acaldas, line 53: Unwanted changes right?
We should only use Redis if a env variable is set, so switchboard works locally

acaldas (approved):

PR #210, opened 2024-06-07
feat: drive by slug

froid1911:


api/src/graphql/server/index.ts
```diff
@@ -57,6 +59,21 @@ export const addGraphqlRoutes = async (
...
+    async (req, res, next) => {
+      const prisma = getExtendedPrisma();
+      const { driveId } = req.params;
+
+      if (!driveId) {
+        throw new Error("driveId required")
+      }
+
+      try {
+        await prisma.document.getDrive(driveId);
```
acaldas, line 71: Since we don't use the value let's use the getDrives query

acaldas (commented):
Let's add this error to the sentry ignore list

acaldas (approved):

PR #247, opened 2024-07-08
feat: set drive icon mutation

froid1911:


api/src/modules/document-drive/drives-resolver.ts
```diff
@@ -113,3 +120,20 @@ export const deleteDrive = mutationField('deleteDrive', {
...
+
+export const setDriveIcon = mutationField('setDriveIcon', {
+  type: 'Boolean',
+  args: {
+    id: nonNull('String'),
+    icon: nonNull('String'),
+  },
+  resolve: async (_parent, { id, icon }, ctx: Context) => {
+    try {
+      await ctx.prisma.document.setDriveIcon(id, icon);
```
acaldas, line 132: the status might not be success, we need to check and throw the associated error

acaldas (approved):

PR #260, opened 2024-07-28
feat: added loki transport for pino

froid1911:


api/src/logger.ts
```diff
@@ -89,6 +90,32 @@ if (process.env.SENTRY_DSN) {
     },
   });
 }
+
+const { LOKI_URL, LOKI_USERNAME, LOKI_PASSWORD } = process.env;
+if (LOKI_URL && LOKI_USERNAME && LOKI_PASSWORD) {
+  const basePath = process.env.BASE_PATH || '/';
+  const baseElements = basePath.split('/');
+
+  const labels = { team: 'powerhouse', application: 'switchboard', env: 'develop' };
```
acaldas, line 99: Should we set the env with LOKI_ENV?

api/src/logger.ts
```diff
@@ -60,11 +61,11 @@ const doesPassFilters = (config: {
 
 const transportTargets: pino.TransportTargetOptions[] = [];
 
-if (isDevelopment) {
```
acaldas, line 63: Is this needed? I'd keep `pino-pretty` active only for dev mode.

acaldas (approved):

PR #263, opened 2024-08-04
feat: added calc for current values

froid1911:


api/src/modules/document/model.ts
```diff
@@ -232,7 +233,10 @@ export function getDocumentDriveCRUD(prisma: Prisma.TransactionClient) {
       driveId: string,
       documentId: string,
     ) => {
-      const document = await driveServer.getDocument(driveId, documentId);
+      let document = await driveServer.getDocument(driveId, documentId);
+      if (document.documentType === 'makerdao/rwa-portfolio') {
```
acaldas, line 237: Let's add a `getDocument` method in `switchboard/api/src/modules/real-world-assets/model.ts`.
Then we call it here inside this `if`

acaldas: Shouldn't they be using the RWA read model?

froid1911: read model isnt live right now

acaldas (approved):

PR #272, opened 2024-08-07
feat: verify signature

froid1911:


api/src/modules/document-drive/drive-resolver.ts
```diff
@@ -408,6 +409,11 @@ export const pushUpdates = mutationField('pushUpdates', {
               branch: 'main'
             })) ?? [];
 
+          if (!(await verifyOperationsAndSignature(operations as Operation<DocumentDriveAction>[]))) { // todo fix this
+            throw new BadRequestError({
```
acaldas, line 413: Should we throw a 401

api/src/modules/document/utils.ts
```diff
@@ -0,0 +1,52 @@
+import { Operation, utils } from "document-model/document";
+import fetch from "node-fetch";
+
+export async function verifyOperationsAndSignature(operations: Operation[]) {
+  const results = await Promise.all(operations.map(async (operation) => {
```
acaldas, line 5: When we move this to the document drive we should try to reuse the hash of an operation as the previousHash for the subsequent operation

api/src/modules/document/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+        );
+      },
+    );
+
+    const data = await fetch(
+      // eslint-disable-next-line max-len
+      `https://renown.id/api/auth/credential?address=${signer.user.address}&chainId=${signer.user.chainId}&app=${signer.app.key}`
+    );
+
+    const credential = await data.json();
```
acaldas, line 42: Renown checks the credential revocation date, but I think it's not checking the expiration date.

api/src/modules/renown/kyc-service.ts
```diff
@@ -0,0 +1,201 @@
...
+      };
+    }>(fetchEntries, {
+      first,
+      skip,
+      input: {
+        and: [
+          {
+            or: [
+              {
+                where: {
```
acaldas, line 51: Here is where we should add a condition for the expirationDate, Renown doesn't have it too

api/src/modules/renown/kyc-service.ts
```diff
@@ -0,0 +1,201 @@
...
+  definition: RuntimeCompositeDefinition
+}
+
+export class KYCService {
+
+  protected client: ComposeClient;
+
+  protected eventSource: EventSource | null = null;
+
+  protected entries: any[] = [];
```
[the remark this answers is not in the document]
acaldas: I'd change entries to credentials to make it clearer.

api/src/modules/renown/kyc-service.ts
```diff
@@ -0,0 +1,201 @@
...
+    this.client = new ComposeClient({
+      ceramic,
+      definition: options.definition as RuntimeCompositeDefinition,
+    });
+
+    this.eventSource = new EventSource(`${ceramic}/api/v0/feed/aggregation/documents`)
+    this.codec = JsonAsString.pipe(AggregationDocument)
+  }
+
+  #fetchEntriesFromCeramic = async (first: number, skip: number) => {
```
acaldas, line 37: Here it would be fetchAllCredentials. So the code calling the kyc-service doesn't need to know that it uses Ceramic.

api/src/modules/renown/kyc-service.ts
```diff
@@ -0,0 +1,201 @@
...
+    const { data } = result;
+    if (!data || data.verifiableCredentialEIP712Index.edges.length === 0) {
+      return null;
+    }
+
+    return data.verifiableCredentialEIP712Index.edges[0]!.node;
+  }
+
+  #updateEntries = async () => {
+    const first = 100;
```
acaldas, line 127: First should be 0 no? Or `this.entries.length`

api/src/modules/renown/kyc-service.ts
```diff
@@ -0,0 +1,201 @@
...
+    let skip = 0;
+    let entries: any[] = [];
+    let foundEntries = 0;
+    const fetchEntriesLoop = async () => {
```
acaldas, line 131: I would use `while` loop instead. Each recursive call is added to the stack, and only when they are all finished are they removed. For the near future there will not be enough credentials to cause a stack overflow but I don't see an advantage in using a recursive call anyway.

I also noticed that recursive calls are harder to debug, when something happens in `processNextJob` on the queues, the callstack is a bunch of `processNextJob` calls which makes it hard to understand the actual caller.

api/src/modules/renown/kyc-service.ts
```diff
@@ -0,0 +1,201 @@
...
+    };
+
+    await fetchEntriesLoop();
+    this.entries = entries;
+  }
+
+  #updateEntry = async (issuerId: string, subjectId: string) => {
+    const entry = await this.#fetchEntryFromCeramic(issuerId, subjectId);
+    const index = this.entries.findIndex((e) => e.issuerId === issuerId && e.subjectId === subjectId);
+    if (!entry && index !== -1) {
```
acaldas, line 150: Ceramic is the source of truth, if it doesn't find the credential we should delete it from our local entries. So here I would remove the `index !== -1` condition.

api/src/modules/renown/kyc-service.ts
```diff
@@ -0,0 +1,201 @@
...
+      delete this.entries[index];
+      return null;
+    }
+
+    if (index !== -1) {
+      this.entries[index] = entry;
```
acaldas, line 156: Same here, if we a different credential stored locally then it should be deleted. Probably the expiration date was reached.

api/src/modules/renown/kyc-service.ts
froid1911: first means the first 100 entries
skip means to skip x entries. we start with 0 and add 100 if there are more pages

api/src/modules/renown/kyc-service.ts
froid1911: I did it this way, because I got a linting error when I used await inside of a while loop.

api/src/modules/renown/kyc-service.ts
froid1911: I just check here whether we store that credential already. if this is the case I overwrite it otherwise I'll add it.
I lookup the index of the entry bei issuerId and subjectId. this means even if there is a new credential. it should overwrite the existing one.

api/src/modules/document/utils.ts
froid1911: could you help me with this?

api/src/modules/document/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+      name: 'ECDSA',
+      namedCurve: 'P-256',
+      hash: 'SHA-256',
+    };
+
+    const verified = await utils.verifyOperationSignature(
+      signer.signatures.at(0)!,
+      signer,
+      async (publicKey, signature, data) => {
+        const importedKey = await crypto.subtle.importKey(
```
froid1911, line 22: this is not working right now :-(

PR #279, opened 2024-08-14
feat: replaced graphql explorer

acaldas:
I've been having some issue with the graphql explorer.
Noticed it hasn't been updated in 2 years and the current main open source explorer seems to be https://github.com/graphql/graphiql.

Screenshot 2024-08-14 at 18 20 57

### DIFF
--- a/api/src/graphql/playground.ts
+++ b/api/src/graphql/playground.ts
@@ -1,0 +1,63 @@
+export function renderGraphqlPlayground(
+  url: string,
+  headers: Record<string, string> = {}
+): string {
+  return `<!doctype html>
+    <html lang="en">
+      <head>
+        <title>GraphiQL</title>
+        <style>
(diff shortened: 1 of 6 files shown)


api/src/graphql/playground.ts
```diff
@@ -0,0 +1,63 @@
+export function renderGraphqlPlayground(
+  url: string,
+  headers: Record<string, string> = {}
+): string {
+  return `<!doctype html>
```
acaldas, line 5: For now, I've just added an html snippet that imports React and Graphqil from [unpkg](https://unpkg.com).
In the future if we have the need to customize the playground we can start bundling it.

acaldas:
Screenshot 2024-08-14 at 18 20 57

froid1911 (approved):
really cool!

PR #293, opened 2024-09-17
feat: added operations query arguments

froid1911:


api/src/modules/document/resolvers.ts
```diff
@@ -40,7 +38,27 @@ export const documentModelInterface = interfaceType({
     t.nonNull.int('revision');
     t.nonNull.field('created', { type: GQLDateBase });
     t.nonNull.field('lastModified', { type: GQLDateBase });
-    t.nonNull.list.nonNull.field('operations', { type: operationModel });
+    t.nonNull.list.nonNull.field('operations', {
+      args: { first: 'Int', skip: 'Int' },
+      type: operationModel, resolve: async (parent, { first, skip }, ctx: Context) => {
+        const operations = ctx.document?.operations;
```
acaldas, line 44: If operations are not defined, it should get them using `ctx.prisma.document.getDocument`

acaldas (approved):

